PR #138, opened 2017-10-12
update restful api

ahutsunshine:


ahutsunshine:
in addition,fix test url path

PR #144, opened 2017-10-23
Add Monitoring Module

xiaoqiu-vip:
user :  xiaoqiu.duan xiaoqiu2017wy@163.com


bhlx3lyx7:
Hi @xiaoqiu-vip , I think the function is great, but in the code there're some hard codes, it's not appropriate for a common project, we recommend the specific literals as configuration, would you please modify it first?
Furthermore, it's better to write comments in English. 
Thanks.

PR #276, opened 2018-05-08
implement sample missing download service, see details in GRIFFIN-117

vzhao:


guoyuepeng:
to download the sample file, where is the specification for the sample? 
say we only download the head 20 records?

vzhao:
good comments!  I will add a API spec here, should download 100 sample records  https://github.com/apache/incubator-griffin/blob/master/griffin-doc/service/api-guide.md

PR #286, opened 2018-05-24
Fix livy parmas escape bug.

nezhazheng:
We escape '`' on SparkSubmitJob, but no unescpae when parse json on org.apache.griffin.measure.Application, so it will be encounter parse error when running job.

### DIFF
--- a/measure/src/main/scala/org/apache/griffin/measure/Application.scala
+++ b/measure/src/main/scala/org/apache/griffin/measure/Application.scala
@@ -39,7 +39,8 @@ object Application extends Loggable {
     }
 
     val envParamFile = args(0)
-    val userParamFile = args(1)
+    // to fix livy bug: character ` will be ignored by livy
+    val userParamFile = args(1).replaceAll("\\\\`", "`")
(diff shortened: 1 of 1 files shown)


bhlx3lyx7:
Hi @nezhazheng , I'll merge it ASAP, thanks for your PR.

guotechfin:
hi nezhazheng,

Could you give us the livy bug id?

Thanks,
William

bhlx3lyx7:
Hi @nezhazheng , thanks for your PR, I've reviewed the code, but it seems to be only in some specific case.
For measure module alone, it should not fix the issues caused by the previous component. 
If livy has such bug, it should fix it, or give some workaround. I'm afraid it's not a common solution here.

bhlx3lyx7:
Hi @nezhazheng , we had also handled that situation in service module before submit to livy [here](https://github.com/apache/incubator-griffin/blob/master/service/src/main/java/org/apache/griffin/core/job/SparkSubmitJob.java#L181), I think in your case you can just add one escape character '\\' before '`', measure module will work well without doing anything.

PR #299, opened 2018-06-14
param json reader test case

grant-xuexu:


bhlx3lyx7:
Hi @grant-xuexu , the file you submit is in lack of the license header, would you add it like the other source files?

PR #305, opened 2018-06-15
refactor params as dq definition

grant-xuexu:


bhlx3lyx7:
Hi @grant-xuexu , there's a conflict file in your PR, would you fix that first? Thanks.

PR #321, opened 2018-06-21
 rename spark.url to resource manager url and log exception stack tree

jasonliaoxiaoge:
rename spark.url to resource manager url and log exception stack tree


guoyuepeng:
LGTM.

guoyuepeng:
rename spark to resource manager is a good proposal, but rm always remind me the remove command, maybe we need to use full name as 'resource.manager', that makes things  clear.

@jasonliaoxiaoge  what do you think?

jasonliaoxiaoge:
Done, change to yarn url

guoyuepeng:
hi Jason,

Thanks, that makes sense.


William

PR #330, opened 2018-06-30
Update griffin_environment.json

shreyansh-sangwan:


guoyuepeng:
hi Shreyansh-sangwan,

Thanks for your updates, but the endpoint could be any port, if you prefer docker you can use 38080.

Thanks,
William

PR #332, opened 2018-07-02
fix a bad practice

toyboxman:
Format string should use %n rather than \n
In format strings, it is generally preferable better to use %n,
which will produce the platform-specific line separator.

### DIFF
--- a/service/src/main/java/org/apache/griffin/core/metric/MetricStoreImpl.java
+++ b/service/src/main/java/org/apache/griffin/core/metric/MetricStoreImpl.java
@@ -81,7 +81,7 @@ public class MetricStoreImpl implements MetricStore {
         this.urlGet = urlBase.concat("/_search?filter_path=hits.hits._source");
         this.urlPost = urlBase.concat("/_bulk");
         this.urlDelete = urlBase.concat("/_delete_by_query");
-        this.indexMetaData = String.format("{ \"index\" : { \"_index\" : \"%s\", \"_type\" : \"%s\" } }\n", INDEX, TYPE);
+        this.indexMetaData = String.format("{ \"index\" : { \"_index\" : \"%s\", \"_type\" : \"%s\" } }%n", INDEX, TYPE);
         this.mapper = new ObjectMapper();
     }
 


whhe:
Hi @toyboxman 

Thanks for your advice, but I think the line separator here has nothing to do with platform. Elasticsearch just use it (in http entity) to identify the bulk request body, see docs : https://www.elastic.co/guide/en/elasticsearch/reference/6.0/docs-bulk.html

Besides, If we use %n here, maybe more code should be modified to keep consistency.
https://github.com/apache/incubator-griffin/blob/37b5cbbe10c7e7c26176a4cfafdceba5ef17e3b1/service/src/main/java/org/apache/griffin/core/metric/MetricStoreImpl.java#L146

toyboxman:
Hi @whhe 

I got you here. but I wanna clarify why I change it.

Java Formatter suggests % as special notation character instead of '\'
https://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html

but if you build customized string like statement you mentioned, '\n' is all right, however line.separator is better.
bulkRequestBody.append("\n");

whhe:
Hi @toyboxman 

I understand your concern and I agree that %n is better than \n as line separator. As you mentioned, the change you committed is acceptable, but not necessary in this situation.

If you insist on changing it, the lines below will be put into the same request body, so we should keep them consistent in structure. 
https://github.com/apache/incubator-griffin/blob/37b5cbbe10c7e7c26176a4cfafdceba5ef17e3b1/service/src/main/java/org/apache/griffin/core/metric/MetricStoreImpl.java#L84
https://github.com/apache/incubator-griffin/blob/37b5cbbe10c7e7c26176a4cfafdceba5ef17e3b1/service/src/main/java/org/apache/griffin/core/metric/MetricStoreImpl.java#L146

And I'm not sure if there is more \n that can be modified, you'b better change them together.

toyboxman:
let me check how many times '\n' was used in codes and make a new patch

guoyuepeng:
hi @toyboxman 
Yes, we might need to new patch for all '\n'

Thanks,
William

toyboxman:
Hi, @whhe @guoyuepeng 

I submit a new patch to revise, please review again

https://github.com/apache/incubator-griffin/pull/335

PR #334, opened 2018-07-03
fix a bad practice, Empty interpolated string

toyboxman:
String declared as interpolated but has no parameters: scala.StringContext.apply("cache after replace old df").s()


guoyuepeng:
I agree 's' Interpolator is unnecessary for these two cases.

PR #343, opened 2018-07-10
GRIFFIN-175 - fix es 411 issue

icesmartjuan:
GRIFFIN-175 -[Service] metric API return 500 - fix es 411 issue


icesmartjuan:
when call get BASE_PATH/api/v1/metrics or get BASE_PATH/api/v1/metrics/values?metricName=jobName&size=5&offset=0, service returns 500. 

after debugging, found that elasticsearch returns 411 Length required response

guoyuepeng:
ok.

PR #344, opened 2018-07-11
GRIFFIN-177 - fix measure page display issue

icesmartjuan:
GRIFFIN-177 - fix measure page display issue


icesmartjuan:
issue:
found UI page wrongly displayed if measure["dq.type"] is not set, 

root cause: <td>row["dq.type"].toLowerCase()</td> in measure.component.html causing infinite loading if  row["dq.type"] is null

PR #346, opened 2018-07-11
GRIFFIN-178 - job state not update after performing start/stop actions on job page

icesmartjuan:
GRIFFIN-178 - job state not update after performing start/stop actions on job page


icesmartjuan:
Issue:
job state is not updated properly after performing start/stop job action, still previous state

PR #353, opened 2018-07-14
GRIFFIN-181-add popup window when performing stop action to a job

icesmartjuan:
GRIFFIN-181-add popup window when performing stop action to a job


icesmartjuan:
add a popup window for user confirmation

PR #354, opened 2018-07-16
GRIFFIN-182 - fix blank flash when routing from job to metric detail

icesmartjuan:
GRIFFIN-182 - fix blank flash when routing from job to metric detail

ISSUE:
currently, when navigating from ```health``` jobs to job-metric detail page, it will reload the health page causing blank flash


icesmartjuan:
other user reported ngx-echarts navigation issue: https://github.com/xieziyu/ngx-echarts/issues/18

PR #366, opened 2018-07-22
GRIFFIN-185 - Feature/download ui

icesmartjuan:
GRIFFIN-185 - Feature/download ui

### DIFF
--- a/service/src/main/java/org/apache/griffin/core/util/FSUtil.java
+++ b/service/src/main/java/org/apache/griffin/core/util/FSUtil.java
@@ -27,7 +27,6 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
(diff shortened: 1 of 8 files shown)


service/src/main/java/org/apache/griffin/core/util/FSUtil.java
```diff
@@ -177,4 +177,23 @@ private static void checkHDFSConf() {
...
     }
 
+    public static String getFirstMissRecordPath(String hdfsDir) throws Exception{
+        List<FileStatus> fileList = listFileStatus(hdfsDir);
+        for(int i=0; i<fileList.size();i++){
+            if(fileList.get(i).getPath().toUri().toString().toLowerCase().contains("missrecord")){
+                return fileList.get(i).getPath().toUri().toString();
+            }
+        }
+        return null;
```
djKooks: Is it okay to return `null`? Because it seems it can be used in `isFileExists` method, and there is no null checking.
Please let me know if I'm thinking wrong. Thanks.

PR #374, opened 2018-07-26
Fix bug and improve code quality

toyboxman:


toyboxman:
invalidate this change and resubmit again later

PR #380, opened 2018-07-31
Fix java import style violations in codes by code style suggested by s…

toyboxman:
…un/google


toyboxman:
@guoyuepeng @bhlx3lyx7 

please consider to keep same code style and we can do those work step by step

PR #382, opened 2018-08-02
Modify measure module to support updated env and config json files format

bhlx3lyx7:
Some important modification in env.json
- "info.cache" -> "griffin.checkpoint"
- "persist" -> "sinks"
- "log" -> "console"
- "http" -> "elasticsearch"
- remove "cleaner"

Some important modiffication in dq.json
- add "baseline" in data source
- add "dataframe.name" in data connector
- in data source,"cache" -> "checkpoint"
- in rule and pre-proc rule, add "in.dataframe.name" and "out.dataframe.name", remove "name"
- in rule, add "out" param array, move "metric", "record" param inside "out" array
- add "sinks" string array as filter of sinks in env.json


guoyuepeng:
reviewed.

We need to update service and ui , and verify on dev branch first.

Thanks,
William

PR #384, opened 2018-08-05
Initialize java code checks rules for griffin

toyboxman:
this config could be used by maven checkStyle plugin


guoyuepeng:
@toyboxman

toyboxman:
I agree with you @guoyuepeng 

actually a code style guide is better than code template here. I will revise this patch

toyboxman:
@guoyuepeng 

I append guide part about how to enable style-check in maven and idea. I think we can add scala and type script code style rule here as well

thanks

PR #392, opened 2018-08-10
Update pom.xml

SparshSinghalHM:
Uncomment Mysql Connector dependency.

### DIFF
--- a/service/pom.xml
+++ b/service/pom.xml
@@ -111,10 +111,10 @@ under the License.
         </dependency>
 
         <!--prod db-->
-        <!--<dependency>-->
-        <!--<groupId>mysql</groupId>-->
-        <!--<artifactId>mysql-connector-java</artifactId>-->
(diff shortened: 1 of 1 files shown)


guoyuepeng:
Hi @SparshSinghalHM 

Thanks for your PR, but we cannot use mysql in our primary pom.xml directly 
since its license not compatible with apache community.

so if you want to use mysql as storage, please use this pom as your build pom

https://github.com/apache/incubator-griffin/blob/master/service/hibernate_mysql_pom.xml

Thanks for your understanding.

William

SparshSinghalHM:
Hi @guoyuepeng 

Can we mention this in the document? 
Also, can we update the document with the latest updates? 

Thanks for responding.

Sparsh

guoyuepeng:
@SparshSinghalHM 

We are updating for our documents.

Thanks,
William

PR #394, opened 2018-08-13
set the fields in configuration classes private

grant-xuexu:
in configuration case classes, get*** functions are defined to access to the fields, so for consistency, make the fields private


guoyuepeng:
hi Grant,

Could you rebase your code before submit PR?

Thanks,
William

PR #396, opened 2018-08-15
set the fields in configuration classes private

grant-xuexu:
in configuration case classes, get*** functions are defined to access to the fields, so for consistency, make the fields private

### DIFF
--- a/measure/src/main/scala/org/apache/griffin/measure/datasource/connector/DataConnectorFactory.scala
+++ b/measure/src/main/scala/org/apache/griffin/measure/datasource/connector/DataConnectorFactory.scala
@@ -91,7 +91,7 @@ object DataConnectorFactory extends Loggable {
                                    ): KafkaStreamingDataConnector  = {
     val KeyType = "key.type"
     val ValueType = "value.type"
-    val config = dcParam.config
+    val config = dcParam.getConfig
     val keyType = config.getOrElse(KeyType, "java.lang.String").toString
     val valueType = config.getOrElse(ValueType, "java.lang.String").toString
     (getClassTag(keyType), getClassTag(valueType)) match {
--- a/measure/src/main/scala/org/apache/griffin/measure/datasource/connector/DataConnectorFactory.scala
+++ b/measure/src/main/scala/org/apache/griffin/measure/datasource/connector/DataConnectorFactory.scala
@@ -94,22 +94,14 @@ object DataConnectorFactory extends Loggable {
     val config = dcParam.getConfig
     val keyType = config.getOrElse(KeyType, "java.lang.String").toString
     val valueType = config.getOrElse(ValueType, "java.lang.String").toString
-    (getClassTag(keyType), getClassTag(valueType)) match {
-      case (ClassTag(k: Class[String]), ClassTag(v: Class[String])) => {
+
+    (keyType, valueType) match {
+      case ("java.lang.String", "java.lang.String") => {
         KafkaStreamingStringDataConnector(sparkSession, ssc, dcParam, tmstCache, streamingCacheClientOpt)
       }
       case _ => {
         throw new Exception("not supported type kafka data connector")
       }
-    }
-  }
-
-  private def getClassTag(tp: String): ClassTag[_] = {
-    try {
-      val clazz = Class.forName(tp)
-      ClassTag(clazz)
-    } catch {
-      case e: Throwable => throw e
     }
   }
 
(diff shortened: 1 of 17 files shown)


grant-xuexu:
"@transient" already is applied to all the child classes

measure/src/main/scala/org/apache/griffin/measure/datasource/connector/DataConnectorFactory.scala
```diff
@@ -91,11 +91,12 @@ object DataConnectorFactory extends Loggable {
...
-    val config = dcParam.config
+    val config = dcParam.getConfig
     val keyType = config.getOrElse(KeyType, "java.lang.String").toString
     val valueType = config.getOrElse(ValueType, "java.lang.String").toString
-    (getClassTag(keyType), getClassTag(valueType)) match {
-      case (ClassTag(k: Class[String]), ClassTag(v: Class[String])) => {
+
+    (keyType, valueType) match {
+      case ("java.lang.String", "java.lang.String") => {
         KafkaStreamingStringDataConnector(sparkSession, ssc, dcParam, tmstCache, streamingCacheClientOpt)
```
bhlx3lyx7: I think match with literal string is too strong with limitation, why not match with the class type?

measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/expr/TreeNode.scala
```diff
@@ -18,28 +18,42 @@ under the License.
...
-    if (this.isInstanceOf[A]) {
-      val tv = seqOp(this.asInstanceOf[A], z)
-      children.foldLeft(combOp(z, tv)) { (ov, tn) =>
-        combOp(ov, tn.preOrderTraverseDepthFirst(z)(seqOp, combOp))
+  def preOrderTraverseDepthFirst[T, A <: TreeNode](z: T)(seqOp: (A, T) => T, combOp: (T, T) => T)(implicit tag: ClassTag[A]): T = {
+
+    val clazz = tag.runtimeClass
+
+    this.getClass match {
+      case `clazz` => {
```
bhlx3lyx7: The code means this.getClass and T must be the same class. But in the usage, it traverses among different types of tree nodes, their concrete classes can't be matched to their parent class, which is not what we want.
I doubt the modification of this file will destroy the process, I suggest you write a unit test for this.

guoyuepeng: we cannot use this patch, please remove this .

guoyuepeng: @bhlx3lyx7 @grant-xuexu we will append a test unit for this.

grant-xuexu: @bhlx3lyx7 yea, you are correct. I misunderstood the logic here. However the issue is `if(this.instanceOf[A])` could already return true because of the type erasure at runtime. 

There is a method called `isAssignableFrom` in Java Class[T], which could be used to check if a class is the subclass of another class. 

so the logic could be changed into:

`
val clazz = tag.runtimeClass

    if(clazz.isAssignableFrom(this.getClass)){
      val tv = seqOp(this.asInstanceOf[A], z)
      children.foldLeft(combOp(z, tv)) { (ov, tn) =>
        combOp(ov, tn.preOrderTraverseDepthFirst(z)(seqOp, combOp))
      }
    }
    else {
      z
    }
`

measure/src/main/scala/org/apache/griffin/measure/datasource/connector/DataConnectorFactory.scala
grant-xuexu: the logic here seems assume the key type and value type are both String, or throw the exception. 

And the type erasure makes `case (ClassTag(k: Class[String]), ClassTag(v: Class[String])) => {` already true

grant-xuexu: Or we could move the type checking logic to configuration validation

bhlx3lyx7: OK, I think you're right, it will be better to match with string here.

bhlx3lyx7:
I think we can merge this PR now.

PR #409, opened 2018-09-03
Support for HDP

SparshSinghalHM:
Added parameter for CISF protection in Ambari. Can be provided in a separate branch.

### DIFF
--- a/service/src/main/java/org/apache/griffin/core/job/SparkSubmitJob.java
+++ b/service/src/main/java/org/apache/griffin/core/job/SparkSubmitJob.java
@@ -63,6 +63,10 @@ import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
 import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestTemplate;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.MediaType;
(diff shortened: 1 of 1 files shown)


service/src/main/java/org/apache/griffin/core/job/SparkSubmitJob.java
```diff
@@ -113,8 +117,13 @@ private void updateJobInstanceState(JobExecutionContext context) throws
     private String post2Livy() {
         String result = null;
         try {
-            result = restTemplate.postForObject(livyUri, livyConfMap,
-                    String.class);
+            HttpHeaders headers = new HttpHeaders();
+            headers.setContentType(MediaType.APPLICATION_JSON);
+            headers.set("X-Requested-By","admin");
```
xuspi: Could you use a constant for the "X-Requested-By"?
  public static final String REQUEST_BY_HEADER = "X-Requested-By";

bhlx3lyx7:
I've tested the code, it works in docker.
"X-Requested-By" could be set as any value, is it appropriate to hardcode it as "admin"?

service/src/main/java/org/apache/griffin/core/job/SparkSubmitJob.java
SparshSinghalHM: Yes. Can do it this way.

SparshSinghalHM:
@bhlx3lyx7  No, not required. We can ask for the ambari username in application.properties. Admin is the default user.

PR #416, opened 2018-09-17
Fix rest api bug in service component

toyboxman:
GET /api/v1/version returns mistaken release version 0.2.0 instead of 0.3.0


toyboxman:
@chemikadze 

I have double checked, release version has been 0.3.0 incubator-griffin/service/src/main/resources/banner.txt.

thank for your reminder

chemikadze:
Oh, sorry, looks like my local copy is out of date already.

chemikadze:
However, looks like GriffinInfoController still should be updated.

PR #421, opened 2018-09-21
GRIFFIN-197 Treat non-existing YARN app as FAILED

chemikadze:
This avoids jobs becoming stuck in UNKNOWN state on Service side.
Also, improves logging for YARN client errors.

### DIFF
--- a/service/src/main/java/org/apache/griffin/core/util/YarnNetUtil.java
+++ b/service/src/main/java/org/apache/griffin/core/util/YarnNetUtil.java
@@ -21,13 +21,16 @@ package org.apache.griffin.core.util;
 
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
-
 import org.apache.commons.lang.StringUtils;
 import org.apache.griffin.core.job.entity.JobInstanceBean;
 import org.apache.griffin.core.job.entity.LivySessionStates;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatus;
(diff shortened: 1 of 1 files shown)


service/src/main/java/org/apache/griffin/core/util/YarnNetUtil.java
```diff
@@ -56,6 +62,14 @@ public static boolean update(String url, JobInstanceBean instance) {
                 instance.setState(LivySessionStates.toLivyState(state));
             }
             return true;
+        } catch (HttpClientErrorException e) {
+            LOGGER.warn("client error {} from yarn: {}",
+                    e.getMessage(), e.getResponseBodyAsString());
+            if (e.getStatusCode() == HttpStatus.NOT_FOUND) {
+                // in sync with Livy behavior, see com.cloudera.livy.utils.SparkYarnApp
+                instance.setState(DEAD);
```
guoyuepeng: Agree we need to handle state, 
but what if this is caused by network issue, 
should we double confirm before we jump to conclusion that the instance is dead?

chemikadze: Only 404 is handled here, which should not be result of network issue.

It looks like any kind of error reported by Yarn client (after internal retries) results in DEADing job on Livy side: https://github.com/cloudera/livy/blob/master/server/src/main/scala/com/cloudera/livy/utils/SparkYarnApp.scala#L307
I'll need to double check whether not found applications are ever getting retried, to make sure behavior is same as on Livy side. If not -- then that's what Livy would do.

PR #428, opened 2018-09-30
Increase build silence timeout

chemikadze:
According to experiment, resolving dependencies takes longer
than default build timeout of 20 minutes. Increasing to 30.

[Experiment](https://travis-ci.org/apache/incubator-griffin/builds/435326777) shown that sometimes it takes more than 20 minutes to download all the dependencies and start executing tests. In that particular experiment it took 22 minutes in order to get to test execution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,8 @@ node_js:
 install:
   - npm install -g bower
 
-script: travis_wait mvn clean verify -q
+script: travis_wait 30 mvn clean verify -q
+
+cache:
+  directories:
+  - $HOME/.m2


chemikadze:
@guoyuepeng as far as I understand, Travis is not preserving state between builds, and maven artifact caches are anyway not cleaned up at clean phase. From other hand, looks like [caching](https://docs.travis-ci.com/user/caching/) can be enabled.

chemikadze:
I've accidentally noticed, that builds are freezing specifically on attempts to access sonatype-apache repository, and no artifacts is successfully downloaded from that repository during the build process. Probably it is worth disabling, too.

toyboxman:
@guoyuepeng @chemikadze 

travis offers cache feature which prevents from downloading maven dependencies every time.

I made pr429 to solve this issue.
https://github.com/apache/incubator-griffin/pull/429

PR #429, opened 2018-10-01
Fix timeout failure in travis CI.

toyboxman:
Build times out because no output was received.

'travis_wait mvn clean verify -q' will spawn a process to deal with mvn task
-q,--quiet     Quiet output - only show errors

-q probably leads no output caught by travis, so timeout failure happens

refer to:
https://docs.travis-ci.com/user/common-build-problems/#limitations-of-travis_wait

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,11 @@ scala:
   - 2.10.6
 node_js:
   - "6"
+git:
+  quiet: true
+cache:
+  directories:
+  - $HOME/.m2
 
 install:
   - npm install -g bower


guoyuepeng:
so that means we don't need to wait 30 minutes?
@toyboxman @chemikadze

chemikadze:
Just removing -q can produce a lot of output. I did few experiments yesterday:

 - without -q: https://travis-ci.org/apache/incubator-griffin/builds/435410581
 - without -q, and "Downloading..." suppressed: https://travis-ci.org/apache/incubator-griffin/builds/435366494
 - without -q and with caching: https://travis-ci.org/apache/incubator-griffin/builds/435428321

Was hesitant introduce another problems with caching or with increasing verbosity, so decided to just bump timeout. "Downloading..." bloats logs a lot and even can fail the build with "[The job exceeded the maximum log length, and has been terminated.](https://travis-ci.org/apache/incubator-griffin/builds/435577065)" error. From other hand travis_wait still would be required if it's suppressed, as resolving dependencies takes most of the time when timeout happens.

Root cause of these timeouts seems to be in repo sonatype-apache responding very slowly, did not find way to disable it yet.

chemikadze:
@guoyuepeng disabling -q hits maximum log length, from other hand suppressing download progress without travis_wait still can hit 10 minute silence timeout

toyboxman:
@guoyuepeng @chemikadze 

I agree to keeping -q parameter, as travis only permit 4M log file size.

please see latest update, decrease git pull/clone log output, I think it should be merged with #428 
https://github.com/apache/incubator-griffin/pull/429/commits/913c9def8a8eafab187ccccc6a6ae1d436beb80d

guoyuepeng:
ok, I will keep both 30m wait time and git quiet option. 
will verify and then might reopen it later.

PR #431, opened 2018-10-07
[GRIFFIN-195] Don't list all tables from UI

chemikadze:
[GRIFFIN-195]
Initially, load only list of databases and list of table names.
Table information is loaded only if user clicks on specific table.

[GRIFFIN-196]
Search by table names.


bhlx3lyx7:
I've tested this PR, the filtering feature works good, but only in profiling creation step. Would you add it in accuracy creation steps as well?

chemikadze:
@bhlx3lyx7 my thoughts exactly, will push in a few minutes

chemikadze:
@bhlx3lyx7 pushed the same for accuracy

bhlx3lyx7:
Cool, seems good to me, I can merge it now.

PR #434, opened 2018-10-10
Griffin-205: accuracy matched fraction

ashutakGG:
https://issues.apache.org/jira/browse/GRIFFIN-205

This pull request covers only batch dq type. 
We need to decide is it worth to add "matched fraction" to streaming type.

Accuracy transformation tests added.


measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/AccuracyExpr2DQSteps.scala
```diff
@@ -125,14 +126,26 @@ case class AccuracyExpr2DQSteps(context: DQContext,
...
-             |(`${totalCountTableName}`.`${totalColName}` - coalesce(`${missCountTableName}`.`${missColName}`, 0)) AS `${matchedColName}`
-             |FROM `${totalCountTableName}` LEFT JOIN `${missCountTableName}`
-         """.stripMargin
+             SELECT `${totalColName}`,
+                    `${missColName}`,
+                    `${matchedColName}`,
+                    coalesce(`${matchedColName}` / `${totalColName}`, 1.0) AS `${matchedFractionColName}`
+
+             FROM (
+               SELECT `${totalColName}`,
```
chemikadze: does it make sense to have aliases for inner selects, and have something like that:
```
`a`.`${totalColName}`
```

measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/AccuracyExpr2DQSteps.scala
```diff
@@ -125,14 +126,26 @@ case class AccuracyExpr2DQSteps(context: DQContext,
...
+                      `${missColName}`,
+                      (`${totalColName}` - `${missColName}`) AS `${matchedColName}`
```
chemikadze: is it possible to compute `matchedFractionColName` here, without extra level of nesting?

measure/src/test/scala/org/apache/griffin/measure/transformations/AccuracyTransformationsIntegrationTest.scala
```diff
@@ -0,0 +1,169 @@
...
+      s"CREATE TABLE IF NOT EXISTS ${EMPTY_PERSON_TABLE} " +
+        "( " +
+        "  name String," +
+        "  age int " +
+        ") " +
+        "ROW FORMAT DELIMITED FIELDS TERMINATED BY ',' LINES TERMINATED BY '\n' " +
+        "STORED AS TEXTFILE"
+    )
+
+    spark.sql(s"select * from ${EMPTY_PERSON_TABLE}").show()
```
chemikadze: assert on emptiness would be useful, what do you think?

measure/src/main/scala/org/apache/griffin/measure/configuration/dqdefinition/DQConfig.scala
```diff
@@ -74,9 +74,9 @@ case class DQConfig(@JsonProperty("name") private val name: String,
   */
 @JsonInclude(Include.NON_NULL)
 case class DataSourceParam( @JsonProperty("name") private val name: String,
-                            @JsonProperty("baseline") private val baseline: Boolean,
```
chemikadze: is that reordering really required?

measure/src/main/scala/org/apache/griffin/measure/configuration/dqdefinition/DQConfig.scala
```diff
@@ -148,12 +148,12 @@ case class EvaluateRuleParam( @JsonProperty("rules") private val rules: List[Rul
...
-                     @JsonProperty("in.dataframe.name") private val inDfName: String,
-                     @JsonProperty("out.dataframe.name") private val outDfName: String,
-                     @JsonProperty("rule") private val rule: String,
-                     @JsonProperty("details") private val details: Map[String, Any],
-                     @JsonProperty("cache") private val cache: Boolean,
-                     @JsonProperty("out") private val outputs: List[RuleOutputParam]
+                     @JsonProperty("in.dataframe.name") private val inDfName: String = null,
+                     @JsonProperty("out.dataframe.name") private val outDfName: String = null,
+                     @JsonProperty("rule") private val rule: String = null,
+                     @JsonProperty("details") private val details: Map[String, Any] = null,
```
chemikadze: is it in sync with json serializer behavior?
how map/list properties are handled if missing from the document?

measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/AccuracyExpr2DQSteps.scala
ashutakGG: Good point. For a sec I even thought you are right (an this is a potential bug). But the following query works as expected.

`spark.sql("SELECT age FROM (SELECT count(age) as age FROM person)").show()`

It show 2 as a result. 
So, looks like high level select does not have an access to tables used for inner select.

But probably I will use this approach to increase readability.

measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/AccuracyExpr2DQSteps.scala
ashutakGG: Whole point of nesting levels here to increase readability. Now, I have an idea where it will "good enough" to read even with 1 level of nesting.

measure/src/main/scala/org/apache/griffin/measure/configuration/dqdefinition/DQConfig.scala
ashutakGG: No, it's not. But keeping fields with default values as last elements in case classes is kind of "best practice" in scala. For example, scala compilation fails for the following snippet of code

```
case class Foo(a: Int, b: Int = 22, c: Double)

val foo = Foo(1, 3.3)
```

For this particular class we can avoid reordering. But I see more reasons to do it right and don't see any reasons to keep it as is (except we are scared to change any pice of code).

measure/src/main/scala/org/apache/griffin/measure/configuration/dqdefinition/DQConfig.scala
ashutakGG: Yes, it exactly is. Actually I expected such question :)
There is a bug in `jackson` support for scala ["Case classes with option types don't follow serialization configuration"](https://github.com/FasterXML/jackson-module-scala/issues/32).
Long story short, It does not support any default values and puts `null`s everywhere. So, if we want to use this classes outside of jackson at the same way as with jackson, `null`s is only chose we have.

measure/src/test/scala/org/apache/griffin/measure/transformations/AccuracyTransformationsIntegrationTest.scala
ashutakGG: yeah. Fixed it with drop tables.

measure/src/main/scala/org/apache/griffin/measure/configuration/dqdefinition/DQConfig.scala
guoyuepeng: LGTM

measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/AccuracyExpr2DQSteps.scala
```diff
@@ -125,14 +126,20 @@ case class AccuracyExpr2DQSteps(context: DQContext,
       // 4. accuracy metric
       val accuracyTableName = ruleParam.getOutDfName()
       val matchedColName = details.getStringOrKey(_matched)
+      val matchedFractionColName = details.getStringOrKey(_matchedFraction)
```
guoyuepeng: @bhlx3lyx7  
@chemikadze 
has this  broken Apache Griffin UI side? Not sure how UI consume from ES.

measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/AccuracyExpr2DQSteps.scala
```diff
@@ -125,14 +126,20 @@ case class AccuracyExpr2DQSteps(context: DQContext,
...
-             |coalesce(`${missCountTableName}`.`${missColName}`, 0) AS `${missColName}`,
-             |(`${totalCountTableName}`.`${totalColName}` - coalesce(`${missCountTableName}`.`${missColName}`, 0)) AS `${matchedColName}`
-             |FROM `${totalCountTableName}` LEFT JOIN `${missCountTableName}`
-         """.stripMargin
+             SELECT A.total AS `${totalColName}`,
+                    A.miss AS `${missColName}`,
+                    (A.total - A.miss) AS `${matchedColName}`,
+                    coalesce( (A.total - A.miss) / A.total, 1.0) AS `${matchedFractionColName}`
+             FROM (
+               SELECT `${totalCountTableName}`.`${totalColName}` AS total,
```
guoyuepeng: this alias LGTM.

measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/AccuracyExpr2DQSteps.scala
```diff
@@ -125,14 +126,20 @@ case class AccuracyExpr2DQSteps(context: DQContext,
...
           s"""
-             |SELECT `${totalCountTableName}`.`${totalColName}` AS `${totalColName}`,
-             |coalesce(`${missCountTableName}`.`${missColName}`, 0) AS `${missColName}`,
-             |(`${totalCountTableName}`.`${totalColName}` - coalesce(`${missCountTableName}`.`${missColName}`, 0)) AS `${matchedColName}`
-             |FROM `${totalCountTableName}` LEFT JOIN `${missCountTableName}`
-         """.stripMargin
+             SELECT A.total AS `${totalColName}`,
+                    A.miss AS `${missColName}`,
+                    (A.total - A.miss) AS `${matchedColName}`,
+                    coalesce( (A.total - A.miss) / A.total, 1.0) AS `${matchedFractionColName}`
```
guoyuepeng: add one more metrics, ok.

measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/AccuracyExpr2DQSteps.scala
bhlx3lyx7: As I know, UI calculates the matched fraction by total and matched count, in ES the metrics are just persisted in JSON format, it's OK to add a field here.

PR #435, opened 2018-10-10
[GRIFFIN-206] fix job timezone when creating from UI

chemikadze:
Demonstration of the problem:
```
> new Date()
Wed Oct 10 2018 11:30:48 GMT-0700 (Pacific Daylight Time)
> new Date().getTimezoneOffset()/60
7
```


chemikadze:
Looks like positive time zones without signs are not parsed properly, so this is not a complete fix.

chemikadze:
@ahutsunshine I've traced time zone logic to your commit, are you aware of problem described in a ticket? Challenge is that UI format is not in sync with parsing logic. Basically there are two options:
 - fix behavior on backend to understand format
 - alternatively we can change UI to use right format
Fixing it on server side might be good (by making existing jobs and examples created with wrong format to be working), from other hand it might break compatibility, as jobs with malformed zones will start to run in new time zone. What's your take on it?

chemikadze:
Went with UI route for a time being, to avoid introducing any incompatible behavior from backend side.

service/src/test/java/org/apache/griffin/core/util/TimeUtilTest.java
```diff
@@ -110,4 +111,36 @@ public void testFormatWithIllegalException() {
         TimeUtil.format(format, time, TimeZone.getTimeZone(timeZone));
     }
 
+    @Test
+    public void testGetTimeZone() {
+        HashMap<String, String> tests = new HashMap<>();
```
ashutakGG: Please, use interface name instead of implementation.

service/src/test/java/org/apache/griffin/core/util/TimeUtilTest.java
```diff
@@ -110,4 +111,36 @@ public void testFormatWithIllegalException() {
...
+        tests.put("GMT1:00", "GMT");
+        tests.put("GMT01:00", "GMT");
+        // values generated by UI for datasets in a past
+        tests.put("UTC1", "GMT");
+        tests.put("UTC1:00", "GMT");
+        tests.put("UTC01:00", "GMT");
+        tests.put("UTC-1", "GMT");
+        tests.put("UTC-1:00", "GMT");
+        tests.put("UTC-01:00", "GMT");
+        for (HashMap.Entry<String, String> e: tests.entrySet()) {
```
ashutakGG: Consider 
```
tests.entrySet().forEach(e -> {...}
```
as an option.

service/src/test/java/org/apache/griffin/core/util/TimeUtilTest.java
```diff
@@ -110,4 +111,36 @@ public void testFormatWithIllegalException() {
...
+            String input = e.getKey();
+            String actual = TimeUtil.getTimeZone(e.getKey()).getID();
```
ashutakGG: Reuse `input` here.

ui/angular/src/app/job/create-job/batch/batch.component.ts
```diff
@@ -134,8 +134,8 @@ export class BatchComponent implements OnInit, AfterViewChecked {
       return false;
     }
     this.measureid = this.getMeasureId();
-    let time = new Date().getTimezoneOffset() / 60;
-    let timezone = "GMT" + time + ":00";
+    let time = -(new Date().getTimezoneOffset() / 60);
```
ashutakGG: Looks like a hack (this and next lines). Probably worth to add a comment. Or even extract common function as long as it happens twice in a code.

ui/angular/src/app/measure/create-measure/configuration/configuration.component.ts
```diff
@@ -56,47 +56,47 @@ export class ConfigurationComponent implements OnInit {
...
+    ["UTC+9(FET)", "GMT+9"],
+    ["UTC+9:30(ACST)", "GMT+9:30"],
+    ["UTC+10(AEST)", "GMT+10"],
+    ["UTC+10:30(FAST)", "GMT+10:30"],
+    ["UTC+11(VTT)", "GMT+11"],
+    ["UTC+11:30(NFT)", "GMT+11:30"],
+    ["UTC+12(PSTB)", "GMT+12"],
+    ["UTC+12:45(CIT)", "GMT+12:45"],
+    ["UTC+13(PSTC)", "GMT+13"],
+    ["UTC+14(PSTD)", "GMT+15"],
```
ashutakGG: 14 -> 15? Is it done intentionally?

ui/angular/src/app/job/create-job/batch/batch.component.ts
ashutakGG: Is there some best practices regarding such logic in angular?

ui/angular/src/app/measure/create-measure/configuration/configuration.component.ts
chemikadze: good catch, thanks

ahutsunshine:
@chemikadze 
You are right. In backend, the method of getTimeZone always returns GMT timezone for  UTC-1 or UTC+01:00 or other similar formats, as the given timezone ID cannot be understood. Currently, I have no better idea to parse different timezone formats in backend. I prefer to solve the problem in front end using a full name such as "America/Los_Angeles", or a standard ID such as "GMT-1:00". Do you have better idea for backend ?

chemikadze:
@ahutsunshine I was thinking on handling UTCxxx format on backend to fix existing jobs, but from other hand, this feels like unnecessary breaking change. I've provided UI fix for now.

service/src/main/java/org/apache/griffin/core/util/TimeUtil.java
```diff
@@ -148,4 +149,11 @@ private static String refreshEscapeHashTag(String str) {
...
     }
 
+    public static TimeZone getTimeZone(String timezone) {
+        if (StringUtils.isEmpty(timezone)) {
+            return TimeZone.getDefault();
+        }
+        return TimeZone.getTimeZone(timezone);
+    }
+
 }
```
guoyuepeng: LGTM, refactor to better place.

ahutsunshine: @chemikadze 
I suggest this code be kept in its original place. Currently just modify ui code to fix bug.

chemikadze: @ahutsunshine Well, it's about time, it's utility function, static makes it easier to test. Felt like it belongs here.

guoyuepeng: yes, it is utility function.

ui/angular/src/app/job/create-job/batch/batch.component.ts
guoyuepeng: IC, this timezone is used for data partition. but can we tell user this critical information in the page?

chemikadze: @guoyuepeng @ashutakGG Moved to utility function. Angular is not my specialty, I was really surprised how much complexity Angular community has about it.

chemikadze:
Added time zone information to batch UI:
screen shot 2018-10-14 at 12 56 43 am

chemikadze:
@ahutsunshine I've found only usage of cron.time.zone is only during schedule creation: https://github.com/apache/incubator-griffin/blob/master/service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java#L423
At the same time I see partition time zone is coming from data.time.zone set in measure JSON: https://github.com/apache/incubator-griffin/blob/master/service/src/main/java/org/apache/griffin/core/job/JobInstance.java#L221 

Am I missing something?

ahutsunshine:
> cron.time.zone

you are right.

chemikadze:
@guoyuepeng @ahutsunshine is there anything left preventing the merge?

guoyuepeng:
let me check

guoyuepeng:
data partition timezone is come from data connector.
https://github.com/apache/incubator-griffin/blob/master/service/src/main/java/org/apache/griffin/core/measure/entity/DataConnector.java#L158

This only relate to scheduler timezone, after tell user this timezone in UI,

I am ok for this now.

Will merge it.

PR #436, opened 2018-10-11
always use apache griffin together as our mark

guoyuepeng:


griffin-doc/dev/dev-env-build.md
```diff
@@ -85,7 +85,7 @@ mvn clean install
 ```
 
 ### For service module and ui module
-1. Login to docker container, and stop running griffin service.
+1. Login to docker container, and stop running apache griffin service.
```
ashutakGG: Probably it's better to use `Apache Griffin` with capital letters everywhere.

guoyuepeng: Agree, will do it.

PR #444, opened 2018-10-24
Define griffin plain-vanilla hook.

toyboxman:
the purpose of hook is for integration with components outside. Griffin would offer information about internal task status.

Task: GRIFFIN-200


service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
```diff
@@ -128,10 +129,22 @@ Licensed to the Apache Software Foundation (ASF) under one
     @Autowired
     private StreamingJobOperatorImpl streamingJobOp;
 
+    @Value("${internal.event.listeners}")
```
chemikadze: Probably it should go to separate `@Configuration`, parametrized by `@Value` and providing `@Bean List<GriffinHook>` or some `@Bean GriffinHookExecutor`. All integration points (JobService, MeasureService) would have this bean autowired.

service/src/main/java/org/apache/griffin/core/integration/GriffinEvent.java
```diff
@@ -0,0 +1,44 @@
...
+ * that registered to receive such events using configuration.
+ *
+ * @author Eugene Liu
+ * @since 0.3
+ */
+public interface GriffinEvent<T> {
+    /**
+     * @return concrete event type
+     */
+    String getType();
```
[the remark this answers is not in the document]
toyboxman: yeah, I have considered this.

if we use enum type, probably we have to change its definition after a new event type is raised.

so I wanna return full qualified class name as correct type name

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
toyboxman: good point, let me change by your suggestion

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
```diff
@@ -168,7 +175,14 @@ public AbstractJob addJob(AbstractJob job) throws Exception {
         Long measureId = job.getMeasureId();
         GriffinMeasure measure = getMeasureIfValid(measureId);
         JobOperator op = getJobOperator(measure.getProcessType());
-        return op.add(job, measure);
+        AbstractJob jobSaved = op.add(job, measure);
+        JobEvent jobEvent = new JobEvent(jobSaved);
+        eventListeners.forEach(hook -> {
```
chemikadze: Feels like passing hooks as list requires all injection points to be aware of execution semantics of hooks. Hiding list of hooks behind some HookChain interface and having `@Bean HookChain` instead of `@Bean List<>` should solve this problem. HookChain would do dispatching of event to all hooks in a list, while all hook uses would be simplified to just `hookChain.onEvent(event)`.

Also I think that in discussions there was an agreement to have blocking execution semantics by default, rather than asynchronous dispatching. To simplify asynchronous hook development, abstract class AsynchronousHook can be provided, which would dispatch calls of abstract onEventAsync method to internal (?) thread pool.

service/src/test/java/org/apache/griffin/core/job/EventServiceTest.java
```diff
@@ -0,0 +1,69 @@
...
+    @Before
+    public void setup() throws Exception {
+        Class.forName("org.apache.griffin.core.integration.JobEventHook");
+        entityManager.clear();
+        entityManager.flush();
+        setEntityManager();
+    }
+
+    @Test
+    public void testAddJobEvent() throws Exception {
```
chemikadze: This test is just verifying that code compiles, and when hook runs, exception is logged. Is that right?

service/src/main/java/org/apache/griffin/core/integration/GriffinEvent.java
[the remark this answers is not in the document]
chemikadze: should it default method on interface be then?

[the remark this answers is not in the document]
guoyuepeng: I think we can use ENUM for open-close principle. 
we close event types for griffin job, and open for downstream listeners.

service/src/main/java/org/apache/griffin/core/integration/GriffinEventListeners.java
```diff
@@ -0,0 +1,61 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+
+@Configuration
+@ConfigurationProperties(prefix = "internal.event")
+public class GriffinEventListeners {
+    private static final Logger LOGGER = LoggerFactory
+        .getLogger(GriffinEventListeners.class);
+    @Value("${internal.event.listeners}")
+    private String listeners;
```
guoyuepeng: We might need to use List<String> as listeners.
In Spring, they will inject string array values as List<String>, as clz1,clz2,clz3

service/src/main/java/org/apache/griffin/core/integration/GriffinEventListeners.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Bean
+    public List<GriffinHook> getListenersRegistered() {
+        ArrayList<GriffinHook> hookList = new ArrayList<>();
+        if (listeners == null || listeners.isEmpty()) {
+            LOGGER.info("Disable griffin event listener for service.");
+        } else {
+            //split list of class name
+            Scanner scanner = new Scanner(listeners).useDelimiter(";");
```
guoyuepeng: Don't need , use spring properties directly.

service/src/main/java/org/apache/griffin/core/integration/JobEvent.java
```diff
@@ -0,0 +1,21 @@
+package org.apache.griffin.core.integration;
+
+import org.apache.griffin.core.job.entity.AbstractJob;
+
+public class JobEvent implements GriffinEvent<AbstractJob> {
```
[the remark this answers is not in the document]
guoyuepeng: from my perspective, I will close all known griffin job events.

service/src/main/java/org/apache/griffin/core/integration/JobEventHook.java
```diff
@@ -0,0 +1,11 @@
+package org.apache.griffin.core.integration;
+
+import org.apache.griffin.core.exception.GriffinException;
+
+public class JobEventHook implements GriffinHook {
+    @Override
+    public void onEvent(GriffinEvent event) throws GriffinException {
+        throw new GriffinException.UnImplementedException(
+            "This method needs to be reimplemented by event-consuming purpose");
```
guoyuepeng: I think griffin will implement and fire to hooks here, right?

service/src/main/java/org/apache/griffin/core/integration/GriffinEventListeners.java
```diff
@@ -0,0 +1,61 @@
...
+            while (scanner.hasNext()) {
+                String hookName = scanner.next();
+                try {
+                    Class<GriffinHook> cl = (Class<GriffinHook>) Class.forName(hookName);
+                    GriffinHook hook = cl.newInstance();
```
guoyuepeng: Internally, can we leverage spring inject hooks for us, 
externally, for deployment time injection, we might use this.
Let's be practical, implement internally way first, make some release, 
then consider externally injection in the future.

toyboxman: I have no feasible/good idea to inject/load external listener now.

so I used java reflection to new listener instance.

let me think about spring inject functionality

service/src/main/java/org/apache/griffin/core/integration/JobEventHook.java
toyboxman: if we think about exporting interface and hope user to implement method details. we should let it empty or throw a exception tip here

service/src/main/java/org/apache/griffin/core/integration/GriffinEventListeners.java
toyboxman: let me try it

service/src/main/java/org/apache/griffin/core/integration/GriffinEventListeners.java
toyboxman: let me try spring list

service/src/test/java/org/apache/griffin/core/job/EventServiceTest.java
toyboxman: yes, I try to catch exception to verify. but I don't find way in spring

service/src/main/java/org/apache/griffin/core/integration/GriffinEvent.java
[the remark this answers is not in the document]
toyboxman: ok, let us change to enum as event type, like

Enum type {
  job,
  measure,
  ...
}

do you agree?

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
toyboxman: in local test, if using synchronous call, listener probably led db transaction failure on condition that listener didn't return.  

I think it's better that injection doesn't impact current code flow

@chemikadze @guoyuepeng  how do you think about

toyboxman:
@guoyuepeng @chemikadze 

Thanks for your comments on first version of hook implementation.

I have completed new implementation, here I simplify hook load, extend event type.

if this is good for you, we can do following work based on this.

best regards

guoyuepeng:
cool!
how about rename GriffinHookChain as GriffinJobEventManager so similar?

toyboxman:
let me change name to GriffinJobEventManage

chemikadze:
Regardig naming -- do you think we should really tie it to Jobs? Handling Measure events could be useful as well.

service/src/main/resources/application.properties
```diff
@@ -62,3 +62,5 @@ elasticsearch.scheme=http
 livy.uri=http://localhost:8998/batches
 # yarn url
 yarn.uri=http://localhost:8088
+# griffin event listener
+#internal.event.listeners=org.apache.griffin.core.event.JobEventHook
```
chemikadze: this example uses outdated format, right?

service/src/test/java/org/apache/griffin/core/job/EventServiceTest.java
chemikadze: `@Test(expected=UnImplementedException.class)`, or `assertThrows()` should do, what do you think?

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
```diff
@@ -158,17 +162,22 @@ public JobServiceImpl() {
...
             throw new GriffinException
-                    .ServiceException("Failed to get RUNNING jobs.", e);
+                .ServiceException("Failed to get RUNNING jobs.", e);
         }
         return dataList;
     }
 
     @Override
     public AbstractJob addJob(AbstractJob job) throws Exception {
+        JobEvent jobEvent = JobEvent.yieldJobEventBeforeCreation(null);
```
chemikadze: shouldn't job be passed here? this would allow to validate some job parameters, for example

service/src/main/java/org/apache/griffin/core/integration/JobEventHook.java
chemikadze: By the way, if this was for testing purposes, we could have test hook defined in test code itself (like in case of hive metadata service tests).

chemikadze:
@toyboxman leaving minor comments aside, looks really nice!

guoyuepeng:
@chemikadze right, we cannot tie events to jobs related. @toyboxman

toyboxman:
summary of comments
1.considering general usage, I do not use 'GriffinJobEventManager' but actually name it 'GriffinEventManager'
2.I keep empty implementation for JobEventHook.onEvent(). because I think it should be done in another change
3.Measure event could follow job event to implement

@guoyuepeng @chemikadze

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
toyboxman: sure, let me change it

toyboxman: @Test(expected=UnImplementedException.class), yeah I ever have considered. but hope not to throw exception in a transaction, so make it empty method.

service/src/main/resources/application.properties
toyboxman: correct, change it as definition in test folder

service/src/main/java/org/apache/griffin/core/event/GriffinEventListeners.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Configuration
+@ConfigurationProperties(prefix = "internal.event")
```
gavlyukovskiy: I'm not sure you need it here, this will map all class fields (that have setters and getters) to properties.

guoyuepeng: right. @toyboxman 
ConfigurationProperties mainly for configuration bean, I see you have used Value, which make sense since you have only one value properties here.
So I agree with @gavlyukovskiy,  I don't think we need support configuration bean here.

toyboxman: thanks guys

I'm gonna remove @ ConfigurationProperties annotation

service/src/test/java/org/apache/griffin/core/job/EventServiceTest.java
guoyuepeng: I think for event test, we need to mockup a event listener in test context, and assert this test listener do accept related event filed from event manager.

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
guoyuepeng: @toyboxman Can we verify it by assert verify on called? instead of some exception throwed.
currently test case , not straightforward to me.

guoyuepeng:
I am ok for registration through declaration in properties.
my problem is can we write triggered on called test case for in EventServiceTests.

toyboxman:
probably we can think about labeling something during onEvent implementation in testcase

toyboxman:
@guoyuepeng 

I revise test case, can you review it again

guoyuepeng:
Based on previous design discuss and code review.
LGTM now.

Any comments from you? @chemikadze

service/src/main/java/org/apache/griffin/core/event/GriffinEventListeners.java
```diff
@@ -0,0 +1,50 @@
...
+    private static final Logger LOGGER = LoggerFactory
+            .getLogger(GriffinEventListeners.class);
+
+    @Autowired
+    private ApplicationContext context;
+    @Value("#{'${internal.event.listeners}'.split(',')}")
+    private List<String> listeners;
+
+    @Bean
+    public List<GriffinHook> getListenersRegistered() {
```
gavlyukovskiy: I see that you getting beans by names from the context, am I right that you're trying to declare only beans that user has specified in the property? I think it will not work, because `@Autowired List<GriffinHook> eventListeners` will always inject all beans of given type from the context.
For example if you do:
```
    @Bean
    public static List<GriffinHook> list() {
        return Arrays.asList(
                new GriffinHook("1"),
                new GriffinHook("2")
        );
    }

    @Bean
    public static GriffinHook hook3() {
        return new GriffinHook("3");
    }

    @Autowired
    private final List<GriffinHook> griffinHooks;
```
in this case spring will try to inject list of beans of given type - list containing only `hook3`.

Instead I would propose to remove this bean declaration and make changes in `GriffinEventManager` to get all beans from context and use only beans specified in the config:
```
public class GriffinEventManager {

    @Autowired
    private ApplicationContext applicationContext;
    @Value("#{'${internal.event.listeners}'.split(',')}")
    private Set<String> enabledListeners;

    private List<GriffinHook> eventListeners;

    @PostConstruct
    void initializeListeners() {
        List<GriffinHook> eventListeners = new ArrayList<>();
        applicationContext.getBeansOfType(GriffinHook.class)
                .forEach((beanName, listener) -> {
                    if (enabledListeners.contains(beanName)) {
                        eventListeners.add(listener);
                    }
                });
        this.eventListeners = eventListeners;
    }

    ...
}
```

gavlyukovskiy:
Left a comment regarding injection. Tests didn't catch it because you explicitly specified all hooks that are registered in the context, I would add one more and leave it unused just for testing purpose.

guoyuepeng:
spring beans framework introspect introspect top-level beans, that is why List<GriffinHook> not wired on.
I agree with @gavlyukovskiy  getbean by name is not as good as getbean by type.
But how to balance between 'extensible' and usable, I am not sure about this.
Do we need to support inspect List<GriffinHook> ?
Can we provide several implements in configuration like GriffinHook1, GriffionHook2... and autowired into List<GriffinHook>?

gavlyukovskiy:
> Can we provide several individually implements in configuration like GriffinHook1, GriffionHook2, and autowired all individual hooks instances into List?
You can define all hooks in the context, but use only ones that explicitly set in property

> Do we really need that much hooks?
We are deploying griffin with custom set of plugins bundled in jar, we will probably define few custom hooks in that jar so that they will be available in griffin context

guoyuepeng:
>> > Can we provide several individually implements in configuration like GriffinHook1, GriffionHook2, and autowired all individual hooks instances into List?
>> 
>> You can define all hooks in the context, but use only ones that explicitly set in property
>> 
>> > Do we really need that much hooks?
>> 
>> We are deploying griffin with custom set of plugins bundled in jar, we will probably define few custom hooks in that jar so that they will be available in griffin context

> IC, users can provide their hooks in application context and griffin somehow need to load and invoke those defined(by-properties) hooks.

gavlyukovskiy:
Yes, I showed this in my example - take all hooks from the context, but use only those that configured in property.

PR #447, opened 2018-10-25
[GRIFFIN-194] single call to fetch all tables of all dbs

chemikadze:
Significantly improves opening speed for measure dialogs.
Previously taking 10+ seconds on huge schemas, with this call
it takes fraction of second to load.

### DIFF
--- a/service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceImpl.java
+++ b/service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceImpl.java
@@ -24,6 +24,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import com.google.common.collect.Lists;
 import org.apache.hadoop.hive.metastore.IMetaStoreClient;
 import org.apache.hadoop.hive.metastore.api.Table;
 import org.slf4j.Logger;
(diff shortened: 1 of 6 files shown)


service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceImpl.java
```diff
@@ -99,8 +100,17 @@ public HiveMetaStoreServiceImpl() {
     public List<Table> getAllTable(String db) {
         return getTables(db);
     }
-
-
+    
+    @Override
+    @Cacheable(unless = "#result==null || #result.isEmpty()")
```
guoyuepeng: ok, add one more service for all table names.

ui/angular/src/app/measure/create-measure/ac/ac.component.ts
```diff
@@ -644,48 +644,37 @@ export class AcComponent implements OnInit, AfterViewChecked {
   }
 
   ngOnInit() {
-    let allDatabases = this.serviceService.config.uri.dblist;
-    let getTableNames = this.serviceService.config.uri.tablenames;
+    let getTableNames = this.serviceService.config.uri.dbtablenames;
```
guoyuepeng: add one more endpoint.

chemikadze:
@guoyuepeng can you please clarify your comments, please? Not very clear what changes are required, if any.

guoyuepeng:
@chemikadze  LGTM. 
Just rushed out for my meeting. sorry for confusing.

PR #448, opened 2018-10-25
[GRIFFIN-208] log exception details when it makes sense

chemikadze:
Sometimes error messages are not descriptive enough and do not allow
to pinpoint exact issue. Exception stack traces should make it easier
to troubleshoot issues.

### DIFF
--- a/measure/src/main/scala/org/apache/griffin/measure/Application.scala
+++ b/measure/src/main/scala/org/apache/griffin/measure/Application.scala
@@ -85,9 +85,11 @@ object Application extends Loggable {
     }
 
     // dq app run
-    dqApp.run match {
-      case Success(_) =>
-        info("process run success")
+    val success = dqApp.run match {
+      case Success(result) =>
+        info("process run result: " + (if (result) "success" else "failed"))
+        result
+
       case Failure(ex) =>
         error(s"process run error: ${ex.getMessage}", ex)
 
@@ -110,6 +112,10 @@ object Application extends Loggable {
     }
 
     shutdown
+
+    if (!success) {
+      sys.exit(-5)
+    }
   }
 
   private def readParamFile[T <: Param](file: String)(implicit m : ClassTag[T]): Try[T] = {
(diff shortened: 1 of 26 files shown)


measure/src/main/scala/org/apache/griffin/measure/Application.scala
```diff
@@ -104,12 +106,18 @@ object Application extends Loggable {
...
       case Failure(ex) =>
-        error(s"process end error: ${ex.getMessage}")
+        error(s"process end error: ${ex.getMessage}", ex)
         shutdown
         sys.exit(-5)
     }
 
     shutdown
+
+    if (success) {
```
chemikadze: looks like this one breaks a lot of stuff

chemikadze: removed exit(0), looks like org.apache.spark.deploy.yarn.ApplicationMaster is not liking direct exit(0)

measure/src/main/scala/org/apache/griffin/measure/Application.scala
```diff
@@ -104,12 +106,16 @@ object Application extends Loggable {
...
+        error(s"process end error: ${ex.getMessage}", ex)
         shutdown
         sys.exit(-5)
     }
 
     shutdown
+
+    if (!success) {
+      sys.exit(-5)
+    }
```
bhlx3lyx7: If any rule step fails, explicitly break down the application can let the spark application shut down with a FAILED state. It makes sense.
Actually, we can think about how to get the job state. Through livy or yarn, we can only get the lifecycle state of spark job, we still need to check logs to address the detailed error, and it would be much more difficult for common users. We can define some job states in calculation, like `starting`, `loading data source`, `pre-processing`, `rule step N success`, and let spark job report its state by configured notify method like http request to service side or something else. In such a way, service can manage states of job instances.

I think job state management is the final solution, explicitly exit the application could work as a temporary solution at current.

PR #453, opened 2018-11-14
Fixservicetests

guoyuepeng:


toyboxman:
LGTM.

it's necessary to rename mock entity class.

PR #457, opened 2018-11-19
merge_pr.py python 3 support

IAmFQQ:
1. Import Request/urlopen/HTTPError from urllib.request to support python 3
2. The print statement has been replaced with a print() function


ludongfang:
LGTM

PR #461, opened 2018-11-27
enhance elasticsearch sink, support elasticsearch security login

justACT:
GRIFFIN-216 [Measure] enhance elasticsearch sink, support elasticsearch security login


guoyuepeng:
@justACT 
could you write two test cases for this?
one is with user/pwd, the other one is without user/pwd.

Thanks,
William

PR #463, opened 2018-12-02
connection issue with user/pwd authentication

justACT:
GRIFFIN-216 [Measure] enhance elasticsearch sink, support elasticsearch security login
 and HttpUtilTest for elasticsearch  with user/pwd and without user/pwd.
maybe this test is not easy to write, for elasticsearch just use http client and I think the " http client" have been tested by its liberary already.

### DIFF
--- a/measure/src/test/scala/org/apache/griffin/measure/utils/HttpUtilTest.scala
+++ b/measure/src/test/scala/org/apache/griffin/measure/utils/HttpUtilTest.scala
@@ -1,0 +1,63 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+*/
+package org.apache.griffin.measure.utils
+
+import org.scalatest.{FlatSpec, Matchers}
+
+class HttpUtilTest extends FlatSpec with Matchers {
+  val url = "http://10.148.181.248:39200/griffin/accuracy"
+  val method = "post"
+  val user = ""
+  val password = ""
+  val params = Map[String, Object]()
+  val header = Map[String, Object](("Content-Type", "application/json"))
+  val data = "{}"
+
+  "postData httpReq" should "be with auth" in {
+    try {
+      HttpUtil.postData(url, user, password, params, header, data)
+    } catch {
+      case e: Throwable => println("cannot connect http client." + e)
+    }
+  }
+
+  "postData httpReq" should "be without auth" in {
+    try {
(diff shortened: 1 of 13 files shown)


measure/src/test/scala/org/apache/griffin/measure/utils/HttpUtilTest.scala
```diff
@@ -0,0 +1,63 @@
...
+  val header = Map[String, Object](("Content-Type", "application/json"))
+  val data = "{}"
+
+  "postData httpReq" should "be with auth" in {
+    try {
+      HttpUtil.postData(url, user, password, params, header, data)
+    } catch {
+      case e: Throwable => println("cannot connect http client." + e)
+    }
+  }
```
bhlx3lyx7: All the test cases are not verified, which is meaningless in unit test.

justACT: but I find if I mock one local variable, I need to use powermock, and it will change pom dependency.

bhlx3lyx7: I don't think there's any static or private function you need to mock, powermock is unnecessary.
`httpReq.postData(data).asString` is not easy to test here, you only need to mock for this line.
BTW, the verification is expected for both success response and parameter existence.

justACT: in scala, how to mock httpReq.postData(data).asString, could you give an example

justACT: or something like when... then return..., not familar with scala test.

bhlx3lyx7: the pattern is not your limit, we're talking about the goal of UT, just to prove your code could accept "username" and "password" and works well.

justACT: ok, got it.

justACT:
maybe it is not easy to mock static class(object), you know that HttpUtil is a Object, https://stackoverflow.com/questions/43475907/mocking-scala-objects-and-functions/43477065 also faces similar problem, if someone is familiar with and interested in it, looking forward your help to finalize this part of UT, thanks.

fusonghe:
Hive tables partition field is by day how to set the condition today and yesterday ratio

guoyuepeng:
> Hive tables partition field is by day how to set the condition today and yesterday ratio

hi fusonghe,

I think you can ask questions in dev mailing list, if your question is not related to this PR.

Send email to dev-subscribe@griffin.apache.org to subscribe 
and then you can discuss your question there.

Thanks,
William

guoyuepeng:
> Hive tables partition field is by day how to set the condition today and yesterday ratio

I didn't understand your question, could you elaborate it or give us an example?

fusonghe:
Why is my mass calculation calculated by the formula to be 0?
image
image
image

fusonghe:
@justACT  Why is my mass calculation calculated by the formula to be 0?

fusonghe:
"mail": {
      "host":"smtp.163.com",
      "mail":"xxx@163.com",
      "user":"xxx@163.com",
      "password":"xxx"
  },
  "sms": {
      "host":"xxx",
      "id":"xx",
      "key":"xxx",
      "UUID":"xxx"
  },
 at dq.json but donot sent mail or messages

bhlx3lyx7:
@justACT I think this PR has been here for too long, with conflicts here and some comments not resolved, can you update the code or just pick part of it as a new PR?

wangmiao1002:
Can the 0.4.0 version be configured to send mail?

PR #480, opened 2019-01-28
For task GRIFFIN-229, trigger the job right now

dyingbleed:


service/src/main/java/org/apache/griffin/core/job/JobController.java
```diff
@@ -113,4 +113,10 @@ public JobHealth getHealthInfo() {
                 .contentType(MediaType.APPLICATION_OCTET_STREAM)
                 .body(resource);
     }
+
+    @RequestMapping(value = "/jobs/trigger/{id}", method = RequestMethod.GET)
```
bhlx3lyx7: The job trigger process is not idempotent, "GET" method seems not proper here. I think "POST" method would be better, as it creates a job submission.

chemikadze:
@dyingbleed That's quite useful feature, and I think many people would really like to see it in upstream. If you don't have time for finishing it, me or @dershovGD can help getting it to completion.

service/src/main/java/org/apache/griffin/core/job/JobController.java
```diff
@@ -113,4 +113,10 @@ public JobHealth getHealthInfo() {
...
+    @ResponseStatus(HttpStatus.NO_CONTENT)
```
chemikadze: I think it makes sense to see how we can return jobinstance information from this method, so it would be possible to wait for job's status via api.

chemikadze:
@bhlx3lyx7 @guoyuepeng We'd like to see GRIFFIN-229 in upstream faster, but original PR author is not responding. What's the policy in this case? Can we take available commit, add commits ontop of it, and send PR, or it should be independent implementation?

guoyuepeng:
@chemikadze @dyingbleed @bhlx3lyx7 

For this feature development, I think we can add commits on top of it.

chemikadze:
Closed by #485

PR #484, opened 2019-03-01
GRIFFIN-232 - add support pluggable predicates

aborgatin:
Classifier "exec" added to spring-boot-maven-plugin for possibility use service as a library. This will allow to create custom predicates without recompiling Griffin-service module. 

The approach to deserialisation in SparkSubmit changed for possibility to use all fields of SegmentPredicate in custom predicates.

### DIFF
--- a/service/pom.xml
+++ b/service/pom.xml
@@ -291,6 +291,7 @@ under the License.
                     </execution>
                 </executions>
                 <configuration>
+                    <classifier>exec</classifier>
                     <fork>true</fork>
                     <layout>JAR</layout>
                     <mainClass>org.apache.griffin.core.GriffinWebApplication</mainClass>
--- a/service/pom.xml
+++ b/service/pom.xml
@@ -44,7 +44,6 @@ under the License.
         <powermock.version>1.6.6</powermock.version>
         <mockito.version>1.10.19</mockito.version>
         <spring-boot-maven-plugin.version>1.5.1.RELEASE</spring-boot-maven-plugin.version>
-        <spring-boot-maven-plugin.classifier></spring-boot-maven-plugin.classifier>
         <derby.version>10.14.1.0</derby.version>
         <eclipselink.version>2.6.0</eclipselink.version>
         <mysql.java.version>5.1.47</mysql.java.version>
--- a/service/pom.xml
+++ b/service/pom.xml
@@ -281,6 +281,23 @@ under the License.
                 </dependencies>
             </plugin>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <version>3.2.1</version>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                            <shadedClassifierName>lib</shadedClassifierName>
+                            <shadedArtifactAttached>true</shadedArtifactAttached>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
                 <groupId>org.springframework.boot</groupId>
                 <artifactId>spring-boot-maven-plugin</artifactId>
                 <version>${spring-boot-maven-plugin.version}</version>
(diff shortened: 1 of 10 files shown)


chemikadze:
@aborgatin can you please add to documentation (probably deployment.md?) how custom segment predicates can be plugged in?

service/src/main/java/org/apache/griffin/core/util/MeasureUtil.java
```diff
@@ -54,6 +55,27 @@ private static void validateGriffinMeasure(GriffinMeasure measure) {
...
+        }
+    }
+
+    private static boolean validatePredicates(GriffinMeasure measure) {
+        for (DataSource dataSource : measure.getDataSources()) {
+            for (DataConnector dataConnector: dataSource.getConnectors()) {
+                for (SegmentPredicate segmentPredicate : dataConnector.getPredicates()) {
+                    try {
+                        Predicator predicator = PredicatorFactory.newPredicateInstance(segmentPredicate);
+                        if (predicator == null) {
```
chemikadze: we throw NotFoundException, so this is not needed

service/src/main/java/org/apache/griffin/core/job/SparkSubmitJob.java
```diff
@@ -185,17 +185,13 @@ private void setPredicates(String json) throws IOException {
         if (StringUtils.isEmpty(json)) {
             return;
         }
-        List<Map<String, Object>> maps = toEntity(json,
-                new TypeReference<List<Map>>() {
+        List<SegmentPredicate> predicates = toEntity(json,
```
chemikadze: is this code covered with tests? if not, let avoid changing any code if not necessary, or provide tests

aborgatin: Yes, this code is covered in tests. I add yet another.

service/src/main/java/org/apache/griffin/core/util/MeasureUtil.java
aborgatin: fixed

aborgatin:
> @aborgatin can you please add to documentation (probably deployment.md?) how custom segment predicates can be plugged in?

I have updated documentation in this part

service/pom.xml
```diff
@@ -291,8 +291,9 @@ under the License.
                     </execution>
                 </executions>
                 <configuration>
+                    <classifier>exec</classifier>
                     <fork>true</fork>
-                    <layout>JAR</layout>
+                    <layout>ZIP</layout>
```
chemikadze: we are modifying existing configuration here -- does it mean that any **existing** artifacts will change their semantics?

chemikadze: will this change be required, if we'll do something like that:
```
java -cp griffin-service.jar -Dloader.path="/path/to/lib/plugin.jar" org.springframework.boot.loader.PropertiesLauncher
```
?

chemikadze:
@aborgatin let keep semantics of existing jar, and build separate "lib" jar to be used as dependency for plugins

service/pom.xml
```diff
@@ -291,7 +292,7 @@ under the License.
                     </execution>
                 </executions>
                 <configuration>
-                    <classifier>exec</classifier>
+                    <classifier>${spring-boot-maven-plugin.classifier}</classifier>
```
chemikadze: Let implement it in a way, that library jar is built on every build with its own classifier ("lib"?), without need to specify classifier explicitly, and keep exec semantics of default jar. Otherwise we'll have semantics of artifacts changing within same classifier depending on build settings, which is very error-prone.

aborgatin: I agree this way is better. I did it, please check.

service/pom.xml
```diff
@@ -280,6 +280,23 @@ under the License.
                     </dependency>
                 </dependencies>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
chemikadze: will the resulting artifact have all dependencies inside? let make sure it does not -- typically "library jar" is meant to have only module's classes, without deps.

aborgatin:
@chemikadze, I replaced the shade plugin with the jar plugin. In this configuration, the “correct” jar is created. Please check

guoyuepeng:
LGTM

PR #485, opened 2019-03-07
GRIFFIN-229 trigger the job right now with fixing comments

aborgatin:
[comment](https://github.com/apache/griffin/pull/480#issuecomment-460075951)
> @dyingbleed can you please include screenshots for UI changes?

screen shot 2019-03-06 at 18 19 25
screen shot 2019-03-06 at 18 43 23


chemikadze:
Looks good! 
@bhlx3lyx7 @dyingbleed WDYT?

chemikadze:
LGTM

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
```diff
@@ -661,4 +654,21 @@ private String getSinksPath(String jsonString) {
             return null;
         }
     }
+
```
dershovGD: It would be nice to have a test on this method

service/src/main/java/org/apache/griffin/core/job/JobController.java
```diff
@@ -113,4 +113,10 @@ public JobHealth getHealthInfo() {
                 .contentType(MediaType.APPLICATION_OCTET_STREAM)
                 .body(resource);
     }
+
```
dershovGD: We have to add this method to our documentation and create new postman examples.

aborgatin: I updated api guide and postman collection

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
aborgatin: done

PR #486, opened 2019-03-10
GRIFFIN-237 Implement service method get JobInstanceBean by id

jhsb25:
Implement service method getting JobInstanceBean by JobInstanceBean id


service/src/main/java/org/apache/griffin/core/job/repo/JobInstanceRepo.java
```diff
@@ -34,6 +34,9 @@ Licensed to the Apache Software Foundation (ASF) under one
 
     JobInstanceBean findByPredicateName(String name);
 
+    @Query("select s from JobInstanceBean s where s.id = ?1")
```
chemikadze: is there an index by `id` field?

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
```diff
@@ -288,6 +288,11 @@ public void deleteJob(String name) throws SchedulerException {
         return updateState(instances);
     }
 
+    @Override
+    public JobInstanceBean findInstance(Long id) {
+        return instanceRepo.findByInstanceId(id);
+    }
```
dershovGD: In case there is no bean with given id, I think, it would be better to throw an exception with explanation. We can log this information as well.

chemikadze: what will happen if it's left as-is?

guoyuepeng:
Could you fix the conflicts?

Thanks,
William

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
dershovGD: I think, it will return null or an exception without any information, what exactly happens.

service/src/main/java/org/apache/griffin/core/job/repo/JobInstanceRepo.java
jhsb25: yes

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
jhsb25: done

jhsb25:
> Could you fix the conflicts?
> 
> Thanks,
> William

jhsb25:
@guoyuepeng can you please approve and/or merge

PR #487, opened 2019-03-13
Bug Fix

toyboxman:
scalaj.http only allows HttpURLConnection to call remote web api,
but on spark cluster condition, no HttpURLConnection returns instead
of class org.apache.hadoop.fs.FsUrlConnection. so rest post api never really calls out,
use httpcomponent to execute rest call.

GRIFFIN-238


toyboxman:
@guoyuepeng, @bhlx3lyx7 , please review and approve this change. because it actually leads ES writing failure.

PR #489, opened 2019-03-18
GRIFFIN-240 Return JobInstanceBean in response to a call Trigger job by id

aborgatin:


griffin-doc/service/api-guide.md
```diff
@@ -549,7 +549,21 @@ curl -k -H "Content-Type: application/json" -H "Accept: application/json" \
 `POST /api/v1/jobs/trigger/{job_id}`
 #### API Example
 ```
-curl -k -X POST http://127.0.0.1:8080/api/v1/jobs/trigger/51
+curl -k -X POST http://127.0.0.1:8080/api/v1/jobs/trigger/51 \
+-d '{
+    "timeout": "1000"
```
chemikadze: what's the purpose of this timeout?

service/src/main/java/org/apache/griffin/core/job/CountDownTriggerListener.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public void triggerFired(Trigger trigger, JobExecutionContext context) {
+        LOGGER.info("CountDownTriggerListener triggerFired");
```
chemikadze: should we add name of trigger or trigger listener here? otherwise, these info log messages are rather useless in production

service/src/main/java/org/apache/griffin/core/job/JobController.java
```diff
@@ -115,8 +120,19 @@ public JobHealth getHealthInfo() {
...
     @RequestMapping(value = "/jobs/trigger/{id}", method = RequestMethod.POST)
-    @ResponseStatus(HttpStatus.NO_CONTENT)
-    public void triggerJob(@PathVariable("id") Long id) throws SchedulerException {
-        jobService.triggerJobById(id);
+    @ResponseStatus(HttpStatus.OK)
+    public JobInstanceBean triggerJob(@PathVariable("id") Long id, @RequestBody(required = false) String request) throws SchedulerException {
+        return jobService.triggerJobById(id, extractTimeOut(request));
+    }
+
+    private long extractTimeOut(String request) {
```
chemikadze: this timeout is rather infra concern, than of particular request, and 99% of all usages won't care about exact value -- let have it as configuration property, not a request body parameter

service/src/main/java/org/apache/griffin/core/job/JobController.java
```diff
@@ -115,8 +120,19 @@ public JobHealth getHealthInfo() {
...
+        long timeout = 0;
```
chemikadze: body is marked as optional, but this will return zero timeout if not provided -- does not look like correct behavior

service/src/main/java/org/apache/griffin/core/job/JobInstance.java
```diff
@@ -281,7 +286,8 @@ private void createJobInstance(Map<String, Object> confMap)
             throw new GriffinException.ConflictException
                     (QUARTZ_JOB_ALREADY_EXIST);
         }
-        saveJobInstance(jobName, groupName);
+        String triggerName = (String) confMap.get(TRIGGER_KEY);
```
chemikadze: why trigger _key_ is named on bean object as trigger _name_?

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
```diff
@@ -666,9 +669,28 @@ public void triggerJobById(Long id) throws SchedulerException {
                     .forJob(jobKey)
                     .startNow()
                     .build();
+            TriggerKey key = trigger.getKey();
+            CountDownLatch latch = new CountDownLatch(1);
+            scheduler.getListenerManager().addTriggerListener(
```
chemikadze: will this listener be automatically deleted later, or should we take care of that?

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
```diff
@@ -666,9 +669,28 @@ public void triggerJobById(Long id) throws SchedulerException {
...
+                    , key::equals);
+
             scheduler.scheduleJob(trigger);
+
+            try {
+                latch.await(timeout, TimeUnit.MILLISECONDS);
+            } catch (InterruptedException e) {
+                LOGGER.warn("CountDownLatch awaiting is interrupted");
+            }
+            List<JobInstanceBean> instanceBeans = instanceRepo.findByTriggerName(key.toString());
```
chemikadze: is it a full scan or not?

service/src/main/java/org/apache/griffin/core/job/entity/JobInstanceBean.java
```diff
@@ -85,6 +85,8 @@ Licensed to the Apache Software Foundation (ASF) under one
     @JsonIgnore
     private AbstractJob job;
 
+    private String triggerName;
```
chemikadze: this will change database schema, isn't it?
will that schema change be applied automatically to database table?
if not, how to migrate database between versions?

griffin-doc/service/api-guide.md
aborgatin: This timeout is the maximum time in milliseconds that we except when creating JobInstanceBean.
The time between schedule job with "execute right now" trigger and the actual execution the job may be differ (from several milliseconds to tens of seconds). 
Sometimes, for example, in requests from the UI, waiting for a JobInstanceBean to be created is not required, because response will be ignored. 
But in other cases it can be very important. For example, we can run job using this API from ETL. After that using JobInstance ID from response, we can check the measurement result.

service/src/main/java/org/apache/griffin/core/job/JobController.java
aborgatin: Why not? The default timeout value is zero, since we do not need to wait for the creation of a JobInstanceBean. If it is important to return a JobInstanceBean, we can set a timeout in the request.

service/src/main/java/org/apache/griffin/core/job/JobController.java
aborgatin: Of course I can remove this param from request and use is only as configuration property. Sometimes there are situations when we need to wait for the creation of JobInstanceBean, but in other situations it is not necessary (see https://github.com/apache/griffin/pull/489#discussion_r266855423)

service/src/main/java/org/apache/griffin/core/job/entity/JobInstanceBean.java
aborgatin: This is schema change will be applied automatically

service/src/main/java/org/apache/griffin/core/job/CountDownTriggerListener.java
aborgatin: I removed it

service/src/main/java/org/apache/griffin/core/job/JobInstance.java
aborgatin: Fixed

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
aborgatin: No, it doesn't happen automatically, I added it

griffin-doc/service/api-guide.md
chemikadze: Please make sure both cases (body is specified, and body is not specified) are covered with tests. Let wait by default (with default specified in config), and specify 0 explicitly, if required.

service/src/main/java/org/apache/griffin/core/job/JobController.java
chemikadze: Got it, makes sense. However, let see why exactly it is taking long time.

service/src/main/java/org/apache/griffin/core/job/JobController.java
chemikadze: Default behavior without waiting could be very confusing for API users, I'm reluctant to have no-wait behavior by default.

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
```diff
@@ -666,9 +668,31 @@ public void triggerJobById(Long id) throws SchedulerException {
...
+            String listenerName = "listenerJob_" + jobKey.toString();
+            try {
+                scheduler.getListenerManager().addTriggerListener(
+                        new CountDownTriggerListener(latch, listenerName)
+                        , key::equals);
+
+                scheduler.scheduleJob(trigger);
+
+                try {
+                    latch.await(0, TimeUnit.MILLISECONDS);
```
chemikadze: why timeout is not used here?

service/src/main/java/org/apache/griffin/core/job/JobController.java
RodionGork: @chemikadze on the other hand, such behavior corresponds to existing behavior, if I got it right. Otherwise it may happen that updated API introduces delays on requests where users perhaps don't expect them yet. Of course it would be amended by default value 0 for timeout in config... Well... I feel like this situation about timeout needs a bit more consideration.

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
```diff
@@ -666,9 +668,31 @@ public void triggerJobById(Long id) throws SchedulerException {
...
+                } catch (InterruptedException e) {
+                    LOGGER.warn("CountDownLatch awaiting is interrupted");
+                }
+                List<JobInstanceBean> instanceBeans = instanceRepo.findByTriggerKey(key.toString());
```
RodionGork: We probably can slightly change this API to become properly asynchronous - we can return triggerKey from the `/trigger` request and then client can send further requests with this key to find the instance as in this line.

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
aborgatin: Yes, it is a full scan

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
aborgatin: Timeout don't used here by mistake, I fixed it

chemikadze:
Superseded by #494

PR #491, opened 2019-03-29
GRIFFIN-244 Get metrics by instance

dershovGD:

PR #492, opened 2019-04-02
GRIFFIN-244 get metrics by instance

RodionGork:
This is just a squashed version of #491

### DIFF
--- a/service/src/main/java/org/apache/griffin/core/metric/MetricStoreImpl.java
+++ b/service/src/main/java/org/apache/griffin/core/metric/MetricStoreImpl.java
@@ -31,6 +31,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.griffin.core.metric.model.MetricValue;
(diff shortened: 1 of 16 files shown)


service/src/main/java/org/apache/griffin/core/metric/model/MetricValue.java
```diff
@@ -61,4 +72,38 @@ public void setTmst(Long tmst) {
     public void setValue(Map<String, Object> value) {
         this.value = value;
     }
+
+    public String getApplicationId() {
```
chemikadze: Suddenly I've realized that probably more extensible way would be to have "metadata" map on metric, instead of fixed "applicationId" value. Who knows what can be useful later.

I know that applicationId is already part of the code base, but as it's not read from anywhere and we did not release that code, it can be right moment to change it. But later we'll be at point of no return.

RodionGork: Ok, I think I got it. So we should add anything which would look in the json like:

    {
        ...
        "metadata": {
            "applicationId": "..."
        }
        ...
    }

Right?

service/src/main/java/org/apache/griffin/core/metric/MetricStoreImpl.java
```diff
@@ -211,9 +213,13 @@ private static String buildBasicAuthString(String user, String password) {
 
     @Override
     public MetricValue getMetric(String applicationId) throws IOException {
-        Map<String, String> map = new HashMap<>();
-        map.put("q", "applicationId:" + applicationId);
-        Response response = client.performRequest("GET", urlGet, map);
+        Response response0 = client.performRequest(
```
dershovGD: Looks like this value never used.

RodionGork: @dershovGD thanks a lot, it was debug stuff. Let me get rid of it...

chemikadze:
@RodionGork let update logic in measure attaching applicationId to metrics in new format

RodionGork:
@chemikadze I've made changes in the measure code according to your hint. I only was not sure if we should change the "external" interface of `MetricWrapper`, but then judged it could be extended whenever we want, so left it as is. Tested the whole stuff locally - and hope to repeat this tomorrow for demo.

c506980

chemikadze:
LGTM!

@apache/griffin-committers can someone else take a look?

guoyuepeng:
LGTM

guoyuepeng:
Will merge this after 0.5.0 is merged into master

guoyuepeng:
hi 
could you rebase and resubmit it again?

Thanks,
William

PR #493, opened 2019-04-02
Griffin 240 trigger

RodionGork:
This PR requires updating tests and postman json yet, but I need it to show the proposed API before moving further. Triggering and searching by triggerKey - both works.


RodionGork:
Sorry, I shall recreate squashed version of this PR for sanity (see #494)

PR #494, opened 2019-04-02
GRIFFIN-240 trigger returns triggerkey

RodionGork:
This PR requires updating tests and postman json yet, but I need it to show the proposed API before moving further. Triggering and searching by triggerKey - both works.

Please look at the changes in `api-guide.md` for the proposed API additions.

This is rework of #489 (in asynchronous way, without timeout)


service/src/main/java/org/apache/griffin/core/job/JobController.java
```diff
@@ -120,8 +124,13 @@ public JobHealth getHealthInfo() {
...
     @RequestMapping(value = "/jobs/trigger/{id}", method = RequestMethod.POST)
-    @ResponseStatus(HttpStatus.NO_CONTENT)
-    public void triggerJob(@PathVariable("id") Long id) throws SchedulerException {
-        jobService.triggerJobById(id);
+    @ResponseStatus(HttpStatus.OK)
+    public Map<String, Object> triggerJob(@PathVariable("id") Long id, @RequestBody(required = false) String request) throws SchedulerException {
+        return Collections.singletonMap("triggerKey", jobService.triggerJobById(id));
+    }
+
+    @RequestMapping(value = "jobs/trigger/{triggerKey:.+}", method = RequestMethod.GET)
```
chemikadze: it would be semantically more correct to have it as different "collection", because ids are completely different

service/src/main/java/org/apache/griffin/core/job/repo/JobInstanceRepo.java
```diff
@@ -53,4 +53,6 @@ Licensed to the Apache Software Foundation (ASF) under one
 
     @Query("select DISTINCT s from JobInstanceBean s where s.state in ?1")
     List<JobInstanceBean> findByActiveState(State[] states);
+
+    List<JobInstanceBean> findByTriggerKey(String triggerKey);
```
chemikadze: let make sure we have all indexes in place

RodionGork: @chemikadze I've added `@Index` annotation, however it will only create index when the table is created - not when it is just modified by adding column. This should be ok, though.

guoyuepeng:
ok

guoyuepeng:
could you rebase and resubmit it again.

RodionGork:
@guoyuepeng & @chemikadze - I've rebased the branches for this PR and for #492 - is it what is supposed, or should I also recreate PR's?

PR #504, opened 2019-07-01
build and run transform steps with multiple threads

wankunde:
There are sequence of dq steps in each DQJob,and run those steps one by one (with foldLeft function).

We can use multiple threads to run some of those steps which have no dependency.

For example:

In a DQBatchJob, a accuracyExpr will have for steps **_missRecords ,_missCount , __totalCount,   accu** .

_missCount and **_totalCount step can run at the same time .

 

In SeqDQStep ,it just need contains some root steps without dependency steps.

If each step knows it's dependency steps, and when they are ready, we can run the step itself .

Running step : 
accu
| |---__missCount
| | |---__missRecords
| |---__totalCount

Running step : 
__missCount
| |---__missRecords

Running step : 
__missRecords

Running step : 
__totalCount

### DIFF
--- a/measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/AccuracyExpr2DQSteps.scala
+++ b/measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/AccuracyExpr2DQSteps.scala
@@ -163,12 +163,11 @@ case class AccuracyExpr2DQSteps(context: DQContext,
         case StreamingProcessType => Nil
       }
 
-      // accuracy current steps
-      val writeSteps1 =
+      val batchWriteSteps =
         accuracyMetricWriteSteps ++ missRecordsWriteSteps ++ missRecordsUpdateWriteSteps
 
       procType match {
-        case BatchProcessType => accuracyTransStep :: writeSteps1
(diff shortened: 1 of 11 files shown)


wankunde:
@guoyuepeng Thanks for your code review. I'm on vacation. And I'll finish this work next week.

wankunde:
@bhlx3lyx7 
Thank you for your advice. 
Yes, may systems have two or more execution engine inside ,and users can change the engine.
But If parallel execution is good than before, and works better, why should we still kept that execution engine?

measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/AccuracyExpr2DQSteps.scala
```diff
@@ -209,8 +208,8 @@ case class AccuracyExpr2DQSteps(context: DQContext,
           }
 
           // extra steps
-          val writeSteps2 = accuracyMetricWriteStep :: accuracyRecordWriteStep :: Nil
-          accuracyRecordTransStep :: writeSteps1 ++ writeSteps2
+          val streamingWriteSteps = accuracyMetricWriteStep :: accuracyRecordWriteStep :: Nil
```
guoyuepeng: LGTM.

PR #506, opened 2019-07-04
add hive metadata jdbc retrieval

qwang6:
Hi,
I used the all default configurations except that changed the database from postgre to mysql.
When ever I added the hive-jdbc dependency in service/pom.xml, I cannot start the Griffin Service normally. 

Error message in attached file.
[griffin_err.log](https://github.com/apache/griffin/files/3360079/griffin_err.log)


qwang6:
What I have changed:
1. add one implementation of HiveMetaStoreService
2. add hive-jdbc dependency in service/pom.xml
3. changed postgre to mysql
4. add griffin server port (default 8080), default application.properties doesn't have this
        server.port = 8080
        spring.application.name=griffin_service

Build:
mvn clean install -DskipTests

Run:
java -jar service/target/service-0.6.0-SNAPSHOT.jar

qwang6:
I have changed get connection code, and now using JDBC to retrieve Hive metadata is working on my local machine.

guoyuepeng:
it seems cannot init cache beans because of 
Caused by: java.lang.ClassNotFoundException: com.google.common.cache.CacheBuilderSpec

could you check your cache part?

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJDBCImpl.java
```diff
@@ -0,0 +1,224 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Author: qwang8
+ * Date:   2019-06-18.
+ */
+@Service
+@Qualifier(value = "hive_jdbc")
+@CacheConfig(cacheNames = "hive", keyGenerator = "cacheKeyGenerator")
```
guoyuepeng: could you rename cache name as jdbchive?

guoyuepeng:
Thanks for your PR.
 
please add test case first
also follow our apache coding standard such as add apache header. 
https://github.com/apache/griffin/blob/master/griffin-doc/dev/code-style.md

Thanks,
William

service/pom.xml
```diff
@@ -167,6 +167,19 @@ under the License.
             </exclusions>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.hive</groupId>
```
guoyuepeng: ok, it is apache license.

qwang6:
1. Have done the code style check
2. Added Unit tests
3. Added Apache License header

guoyuepeng:
https://travis-ci.org/apache/griffin/builds/554517265?utm_source=github_status&utm_medium=notification

Tests in error: 
  HiveMetaStoreControllerTest.testGetAllDatabases » IllegalState Failed to load ...
  HiveMetaStoreControllerTest.testGetAllTableNames » IllegalState Failed to load...
  HiveMetaStoreControllerTest.testGetAllTables » IllegalState Failed to load App...
  HiveMetaStoreControllerTest.testGetAllTablesWithDb » IllegalState Failed to lo...
  HiveMetaStoreControllerTest.testGetTable » IllegalState Failed to load Applica...
  MetricStoreImplTest.testMetricGetting:110 » IllegalArgument Host name may not ...
Tests run: 193, Failures: 0, Errors: 6, Skipped: 0


Did you forget to submit related test cases?

guoyuepeng:
hi,

to fix the test case error, you might need to separate it into two test cases for HiveMetaStoreControllerTest, one is using metastore , the other cases is to jdbc as below.

@RunWith(SpringRunner.class)
@WebMvcTest(value = HiveMetaStoreController.class, secure = false)
public class HiveMetaStoreControllerTest {

    @Autowired
    private MockMvc mockMvc;

    @MockBean
    @Qualifier(value = "hive_jdbc")
    private HiveMetaStoreService hiveMetaStoreService;



Thanks,
William

service/pom.xml
```diff
@@ -37,7 +37,7 @@ under the License.
         <hadoop.version>2.7.1</hadoop.version>
         <hive.version>2.2.0</hive.version>
         <scala.version>2.10</scala.version>
-        <spring.boot.version>1.5.1.RELEASE</spring.boot.version>
+        <spring.boot.version>1.5.10.RELEASE</spring.boot.version>
```
guoyuepeng: please don't change version until you have specific reason.

service/pom.xml
```diff
@@ -105,16 +105,16 @@ under the License.
...
+<!--        <dependency>-->
+<!--            <groupId>org.postgresql</groupId>-->
+<!--            <artifactId>postgresql</artifactId>-->
+<!--            <version>${postgresql.version}</version>-->
+<!--        </dependency>-->
         <dependency>
-            <groupId>org.postgresql</groupId>
-            <artifactId>postgresql</artifactId>
-            <version>${postgresql.version}</version>
+            <groupId>mysql</groupId>
```
guoyuepeng: please don't change the default db.

service/pom.xml
```diff
@@ -231,31 +244,31 @@ under the License.
...
-            <!--<id>mysql</id>-->
-            <!--<activation>-->
-                <!--<property>-->
-                    <!--<name>mysql</name>-->
-                <!--</property>-->
-            <!--</activation>-->
-        <!--</profile>-->
         <profile>
-            <id>dev</id>
+            <id>mysql</id>
```
guoyuepeng: please don't change the default profile.

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
```diff
@@ -160,6 +143,22 @@ public void evictHiveCache() {
         LOGGER.info("Evict hive cache");
     }
 
+    private void closeConnection(Connection conn, Statement stmt, ResultSet rs) {
+        try {
+            if (rs != null) {
```
guoyuepeng: I see these three resources are AutoCloseable. please use try-with-resource. don't need to close them in finally block.
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreController.java
```diff
@@ -34,6 +35,7 @@ Licensed to the Apache Software Foundation (ASF) under one
 public class HiveMetaStoreController {
 
     @Autowired
+    @Qualifier(value = "hive_metastore")
```
ahutsunshine: Pls use lowerCamelCase to name variables. The following code is also the same.

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    public void setHiveClassName(String hiveClassName) {
+        this.hiveClassName = hiveClassName;
+    }
+
+    @Override
+    @Cacheable(unless = "#result==null")
+    public Iterable<String> getAllDatabases() {
+        String sql = "show databases";
```
ahutsunshine: Perhaps it would be better to abstract it into a static final constant, like
`public static final String SHOW_DATABASES = "show databases";`
The following code is also.

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
```diff
@@ -0,0 +1,287 @@
...
+
+        try {
+            Class.forName(hiveClassName);
+            if (conn == null) {
+                conn = DriverManager.getConnection(hiveUrl);
+            }
+        } catch (ClassNotFoundException | SQLException e) {
+            e.printStackTrace();
+        }
+
```
ahutsunshine: 1.Pls use `LOGGER.error ` instead of `e.printStackTrace()`
2.If exception occurs, here just catches it and doesn't do anything. Also, the next query is still to report exceptions. 
May be register class name and hive connection be done after class created, like 
```@PostConstruct
    public void init() {
        Class.forName(hiveClassName);
        if (conn == null) {
            conn = DriverManager.getConnection(hiveUrl);
        }
    }
```
you can also add retry to connect hive.
This is just my advice.

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
```diff
@@ -0,0 +1,287 @@
...
+        if (s == -1 || e == -1) {
+            return "";
+        }
+        if (s > e) {
+            return "";
+        }
+
+        return colStr.substring(s + 1, e);
+    }
+}
```
ahutsunshine: Can we use regex to match ' ' and get column comment?  This seems not easy to understand.

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
```diff
@@ -0,0 +1,287 @@
...
+
+        return tableMetadata.substring(start + 1, end);
+    }
+
+    /**
+     * Get the Hive table schema: column name, column type, column comment
+     * @param tableMetadata hive table metadata string
+     * @return List of FieldSchema
+     */
+    public List<FieldSchema> getColums(String tableMetadata) {
```
ahutsunshine: Can you give an example for tableMetadata in getColums() Javadoc?

service/src/main/resources/application.properties
```diff
@@ -27,6 +28,11 @@ hive.metastore.uris=thrift://localhost:9083
 hive.metastore.dbname=default
 hive.hmshandler.retry.attempts=15
 hive.hmshandler.retry.interval=2000ms
+
+#Hive jdbc
+hive.jdbc.className=org.apache.hive.jdbc.HiveDriver
+hive.jdbc.url=jdbc:hive2://localhost:10000/
+
```
ahutsunshine: Can you help add a comment and give an example to introduce Kerberos authentication for hive?

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
qwang6: In current version I used the DriverManager to get Hive Connection, if changed to try-with-resource, I cannot do the unit test with Mockito. So we keep try-catch-finally for now, and if we will change to use Hive Connection Pool, we may change to try-with-resource method.

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJDBCImpl.java
qwang6: done

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreController.java
qwang6: done

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
qwang6: done

service/src/main/resources/application.properties
qwang6: In this PR, I didn't add Hive Kerberos auth.

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
qwang6: done

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
qwang6: done

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
qwang6: It's not easy to use Mockito to do the unit test in this way. I will try to use Hive connection pool to replace this solution in future.

guoyuepeng:
I have reviewed all changes.

Currently we have many env related settings, such as 
db
 - mysql
 - postgresql
auth
 - noauth
 - keytab
metadata
 - jdbc connect directly
 - hive
etc,.


Let us think how to uniformly handle modules to separate public simple deployment vs enterprise specific deployment.

Thanks,
William

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
```diff
@@ -0,0 +1,344 @@
...
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+*/
+
+package org.apache.griffin.core.metastore.hive;
+
+import java.io.IOException;
+import java.sql.*;
```
toyboxman: @qwang6 ,

please change your ide setting and auto import classes used instead of import asterisk

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
```diff
@@ -0,0 +1,344 @@
...
+        if (needKerberos != null && needKerberos.equalsIgnoreCase("true")) {
+            LOGGER.info("Hive need Kerberos Auth.");
+
+            Configuration conf = new Configuration();
+            conf.set("hadoop.security.authentication", "Kerberos");
+            UserGroupInformation.setConfiguration(conf);
+            try {
+                UserGroupInformation.loginUserFromKeytab(keytabUser, keytabPath);
+            } catch (IOException e) {
+                e.printStackTrace();
```
toyboxman: please avoid default exception handling

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
```diff
@@ -0,0 +1,344 @@
...
+    @Override
+    @Cacheable(unless = "#result==null")
+    public Iterable<String> getAllDatabases() {
+        return queryHiveString(SHOW_DATABASE);
+    }
+
+    @Override
+    @Cacheable(unless = "#result==null")
+    public Iterable<String> getAllTableNames(String dbName) {
+
```
toyboxman: remove this blank line

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
```diff
@@ -0,0 +1,344 @@
...
+        return queryHiveString(SHOW_TABLES_IN + dbName);
+    }
+
+    @Override
+    @Cacheable(unless = "#result==null")
+    public Map<String, List<String>> getAllTableNames() {
+//        // Not recommend this method because this method will generate one query for every database
```
toyboxman: if the commented codes are useless, you should remove them from formal codes

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
```diff
@@ -0,0 +1,344 @@
...
+        Statement stmt = null;
+        ResultSet rs = null;
+        StringBuilder sb = new StringBuilder();
+
+        try {
+            Class.forName(hiveClassName);
+            if (conn == null) {
+                conn = DriverManager.getConnection(hiveUrl);
+            }
+            LOGGER.info("got connection");
```
toyboxman: it's better add hiveUrl connection into log output

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
```diff
@@ -0,0 +1,344 @@
...
+     * Query Hive for Show tables or show databases, which will return List of String
+     * @param sql sql string
+     * @return
+     */
+    private Iterable<String> queryHiveString(String sql) {
+        List<String> res = new ArrayList<>();
+        Statement stmt = null;
+        ResultSet rs = null;
+
+        try {
```
toyboxman: would you like to consider refactoring db access code snippet into a common method

service/pom.xml
```diff
@@ -32,14 +32,12 @@ under the License.
     <name>Apache Griffin :: Web Service</name>
 
     <properties>
-        <java.version>1.8</java.version>
```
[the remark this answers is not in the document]
qwang6: I didn't change java basic config in pom.xml

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
qwang6: #Hive jdbc
hive.jdbc.className=org.apache.hive.jdbc.HiveDriver
hive.jdbc.url=jdbc:hive2://localhost:10000/
hive.need.kerberos=false
hive.keytab.user=xxx@xx.com
hive.keytab.path=/path/to/keytab/file

It's added in application.properties in this pr

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
qwang6: Cannot remove because the formal codes call this method. It's ok for reading from Hive metastore, but will have connection issue when reading from Hive JDBC, especially Hive have a lots of databases and tables.

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
qwang6: I know it's not an elegant solution, but is a working solution. I have plan to change the db access code in future.

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreServiceJdbcImpl.java
ahutsunshine: > In current version I used the DriverManager to get Hive Connection, if changed to try-with-resource, I cannot do the unit test with Mockito. So we keep try-catch-finally for now, and if we will change to use Hive Connection Pool, we may change to try-with-resource method.

I also suggest you to use try-with-resource. For test, I think there is a way to do it .

qwang6:
close this pr and create a clean one

PR #507, opened 2019-07-08
GRIFFIN-263 Use independent logger for griffin

wankunde:
When we run DQJob,  user can only set logger level for spark execution engine. 
```java
  sparkSession.sparkContext.setLogLevel(sparkParam.getLogLevel)
```

We have independent logger for griffin, which would be very helpful to debug or understand the internal of griffin. 
```java
var logLevel = getGriffinLogLevel() sparkSession.sparkContext.setLogLevel(sparkParam.getLogLevel) griffinLogger.setLevel(logLevel)
```


guoyuepeng:
LGTM

PR #514, opened 2019-07-23
[GRIFFIN-269] Griffin submit job to livy server with Kerberos Authentication

qwang6:
In production environment, submit to Livy Server usually using Kerberos Auth. Add three properties to check whether using Kerberos Auth or not.

livy.need.kerberos=false
livy.server.auth.kerberos.principal=livy/kerberos.principal
livy.server.auth.kerberos.keytab=/path/to/livy/keytab/file


service/src/main/java/org/apache/griffin/core/job/SparkSubmitJob.java
```diff
@@ -19,55 +19,40 @@ Licensed to the Apache Software Foundation (ASF) under one
...
 import org.apache.griffin.core.measure.entity.GriffinMeasure.ProcessType;
 import org.apache.griffin.core.util.JsonUtil;
-import org.quartz.DisallowConcurrentExecution;
-import org.quartz.Job;
-import org.quartz.JobDetail;
-import org.quartz.JobExecutionContext;
-import org.quartz.PersistJobDataAfterExecution;
-import org.quartz.SchedulerException;
-import org.quartz.SimpleTrigger;
+import org.quartz.*;
```
toyboxman: please use explicit import instead of asterisk

FYI
https://github.com/apache/griffin/blob/master/griffin-doc/dev/code-style.md

service/src/main/java/org/apache/griffin/core/job/JobServiceImpl.java
```diff
@@ -527,7 +525,9 @@ private void syncInstancesOfJob(JobInstanceBean instance) {
             new TypeReference<HashMap<String, Object>>() {
             };
         try {
-            String resultStr = restTemplate.getForObject(uri, String.class);
+            String resultStr = livyTaskSubmitHelper.getFromLivy(uri);
```
guoyuepeng: ok, it supported kerberos

guoyuepeng:
LGTM.

service/src/main/java/org/apache/griffin/core/job/SparkSubmitJob.java
qwang6: Done. This should be the IDE automatically changed to asterisk.

service/src/main/java/org/apache/griffin/core/job/LivyTaskSubmitHelper.java
```diff
@@ -216,8 +222,105 @@ public void decreaseCurTaskNum(Long scheduleId) {
...
+
+        HttpHeaders headers = new HttpHeaders();
+        headers.setContentType(MediaType.APPLICATION_JSON);
+        headers.set(REQUEST_BY_HEADER,"admin");
+
+        if (needKerberos == null || needKerberos.isEmpty()) {
+            logger.error("The property \"livy.need.kerberos\" is empty");
+            return null;
+        }
+
```
ahutsunshine: If livy.need.kerberos is empty,  I think users may not need Kerberos. In other words, it's the same as  `livy.need.kerberos = false`.

service/src/main/java/org/apache/griffin/core/job/LivyTaskSubmitHelper.java
```diff
@@ -216,8 +222,105 @@ public void decreaseCurTaskNum(Long scheduleId) {
...
+            String keyTabLocation = env.getProperty("livy.server.auth.kerberos.keytab");
+            logger.info("principal:{}, lcoation:{}", userPrincipal, keyTabLocation);
+
+            KerberosRestTemplate restTemplate = new KerberosRestTemplate(keyTabLocation, userPrincipal);
+            HttpEntity<String> springEntity = null;
+            try {
+                springEntity = new HttpEntity<>(toJsonWithFormat(livyConfMap), headers);
+            } catch (JsonProcessingException e) {
+                e.printStackTrace();
+            }
```
ahutsunshine: Pls use LOGGER.error and return while catching exception.

ahutsunshine:
ok, I see this pr is closed. Later we can optimize this.

PR #518, opened 2019-07-31
[GRIFFIN-273] Griffin Service with user-defined log4j2 configuration 

joohnnie:
Currently, the Griffin service module use spring-boot default log4j2 configuration.

This approach is not easy to change the config for debugging and add new appender such as 
RollingFile


guoyuepeng:
hi Joohnnie, 
Could you rebase, some commits are already merged before.

joohnnie:
Hi William,
I have rebased and please help to check.

Thanks
Johnnie

guoyuepeng:
Thanks, Johnnie.

whhe:
It seems that your configuration is based on the default configuration of spring boot, I think there is no problem with the content. 

But I'm not sure if giving an implementation like what you do or giving a guide in a document is better, cause If users want customized log, they may still need to make some changes in these files.

joohnnie:
@whhe Thanks for your comments. 
The config is the default configuration from spring boot. The reason I put it to the code base because we can easily change the config based on our need. Otherwise, users even don't know what' the log library they used.
In addition to that, I removed some dependencies as they will cause the confliction in the log system.

Regarding the guild, log4j2 is an open-source library and has its official website.
User can refer to the website to customize the config.  
https://logging.apache.org/log4j/2.x/index.html

service/pom.xml
```diff
@@ -82,6 +82,10 @@ under the License.
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
```
guoyuepeng: LGTM.

guoyuepeng:
LGTM.

PR #520, opened 2019-08-09
[GRIFFIN-276]Add batch type measure job test case

wankunde:
Add some test case for batch type measure job, which is helpful for checking the correctness of the code and troublethooting.


guoyuepeng:
Thanks, a good contribution.
Let me think the whole test issue.

measure/src/main/scala/org/apache/griffin/measure/launch/DQApp.scala
```diff
@@ -32,6 +33,8 @@ trait DQApp extends Loggable with Serializable {
   val envParam: EnvConfig
   val dqParam: DQConfig
 
+  implicit var sparkSession: SparkSession = _
```
guoyuepeng: make sense to me.

measure/src/test/scala/org/apache/griffin/measure/job/BatchDQAppTest.scala
```diff
@@ -0,0 +1,155 @@
...
+    // check Result Metrics
+    val dqContext = dqApp.asInstanceOf[BatchDQApp].dqContext
+    val timestamp = dqContext.contextId.timestamp
+    val expectedMetrics =
+      Map(timestamp -> metrics)
+
+    dqContext.metricWrapper.metrics should equal(expectedMetrics)
+  }
+
+  "accuracy batch job" should "work" in {
```
guoyuepeng: GOOD Test case for different data quality dimensions.

guoyuepeng:
LGTM

PR #521, opened 2019-08-14
GRIFFIN-278 AvroBatchDataConnector handle input is directory

joohnnie:
AvroBatchDataConnector process data based on file-level and need to handle the case when input is a directory.


guoyuepeng:
LGTM

PR #524, opened 2019-08-20
[GRIFFIN-281] Add custom sink testcase

wankunde:
Add custom sink test.

I also have two question:
1. In `ConsoleSink`,`ElasticSearchSink`,`MongoSink` , the `sinkRecords` method doesn't really sink records ? 
2.  Why should we have four metrics output types ? 

```
* flatten: Aggregation method used before sending data frame result into the sink:  
      - default: use "array" if data frame returned multiple records, otherwise use "entries" 
      - entries: sends first row of data frame as metric results, like like `{"agg_col": "value"}`
      - array: wraps all metrics into a map, like `{"my_out_name": [{"agg_col": "value"}]}`
      - map: wraps first row of data frame into a map, like `{"my_out_name": {"agg_col": "value"}}`
  
```


guoyuepeng:
LGTM

toyboxman:
LGTM

PR #526, opened 2019-08-22
[GRIFFIN-283] Move sink steps into TransformStep

wankunde:
Treat sink steps as a part of a transform step, so we can keep focus on transform step codes.
Also the sink steps and some other transform step could be executed concurrently.

### DIFF
--- a/measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/Expr2DQSteps.scala
+++ b/measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/Expr2DQSteps.scala
@@ -21,9 +21,10 @@ package org.apache.griffin.measure.step.builder.dsl.transform
 import org.apache.griffin.measure.Loggable
 import org.apache.griffin.measure.configuration.dqdefinition.RuleParam
 import org.apache.griffin.measure.configuration.enums._
-import org.apache.griffin.measure.context.{ContextId, DQContext, TimeRange}
+import org.apache.griffin.measure.context.DQContext
 import org.apache.griffin.measure.step.DQStep
 import org.apache.griffin.measure.step.builder.dsl.expr.Expr
+import org.apache.griffin.measure.step.write.{MetricWriteStep, RecordWriteStep, WriteStep}
 
 trait Expr2DQSteps extends Loggable with Serializable {
 
@@ -31,7 +32,6 @@ trait Expr2DQSteps extends Loggable with Serializable {
   protected val emptyMap = Map[String, Any]()
 
   def getDQSteps(): Seq[DQStep]
-
 }
 
 /**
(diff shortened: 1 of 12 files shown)


measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/Expr2DQSteps.scala
```diff
@@ -32,6 +33,10 @@ trait Expr2DQSteps extends Loggable with Serializable {
 
   def getDQSteps(): Seq[DQStep]
 
+  implicit def recoredWriteStepToList(writeStep: RecordWriteStep): List[WriteStep] = writeStep :: Nil
+
+  implicit def metricsWriteStepToList(writeStep: MetricWriteStep): List[WriteStep] = writeStep :: Nil
```
bhlx3lyx7: It's a little weird to implicitly change a `WriteStep` into `List[WriteStep]`, it would be clearer and easier for reading if we just pass the List parameter as the function defined.

guoyuepeng:
accu
|   |---__totalCount
|   |---__missCount
|   |   |---__missRecords

for accuracyTransStep, 
it has two pre-required steps, missCountTransStep and totalCountTransStep.
it has list of writing subsequent steps, as writesteps (metricwritestep::Nil).

I agree with you to chain those steps.
it is clear.
@bhlx3lyx7 @wankunde 
to be clear to griffin community, maybe we also can refactor step as task in all measure.

Thanks,
William

measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/Expr2DQSteps.scala
guoyuepeng: I also think implicitly convert WriteStep into List[WriteStep] is over-assumption. we should explicitly declare it when used. 
@wankunde What is your opinion?
If we have to change the interface, then we can change it to make dq chain more clear.

wankunde: @bhlx3lyx7 @guoyuepeng 
At first , I think for a TransformStep there could be one sink or a sink list. So I add a implicit method，and the argument can be WriteStep or List[WriteStep].

But now,I think for the result table or dataframe of a transformStep, there could be only one WriteStep( one sink can sink to different places  and we don't need different output logic). So  i want to chuange the interface to **Option[WriteStep]** .

Is that ok ?

guoyuepeng: make sense to me .

measure/src/main/scala/org/apache/griffin/measure/step/transform/DataFrameOpsTransformStep.scala
```diff
@@ -19,14 +19,16 @@ under the License.
...
 
 /**
   * data frame ops transform step
   */
-case class DataFrameOpsTransformStep(name: String,
+case class DataFrameOpsTransformStep[T <: WriteStep](name: String,
                                      inputDfName: String,
                                      rule: String,
                                      details: Map[String, Any],
+                                     writeStepOpt: Option[T] = None,
```
guoyuepeng: LGTM

PR #531, opened 2019-09-05
[GRIFFIN-286]Remove spark-testing-base dependency jar

wankunde:
Now we use spark-testing-base jar to test spark job in measure module, but this jar maybe conflict with the spark version(CDH spark version,spark AE) or scala version(few scala version with specified spark version). 

So I suggest removing the dependency of this package.


guoyuepeng:
LGTM

PR #532, opened 2019-09-05
[GRIFFIN-287]Use Jackson to parse json data

wankunde:
When I build griffin with cdh's spark, the project compile failed because it can not find org.json:json jar in it's dependency tree.

So could we parse json data with jackson or some else jar instead of org.json:json ?

### DIFF
--- a/service/pom.xml
+++ b/service/pom.xml
@@ -96,6 +96,11 @@ under the License.
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.json</groupId>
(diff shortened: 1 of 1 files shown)


wankunde:
Spring-boot read configurations using org.json:json jar.

guoyuepeng:
https://travis-ci.org/apache/griffin/builds/581180967?utm_source=github_status&utm_medium=notification

memory issue?

wankunde:
@guoyuepeng Griffin will compile error with cdh spark while success with apache spark. I will compare  the dependency tree later.

PR #534, opened 2019-09-12
[GRIFFIN-290]Fix bug for submitting job to livy

wankunde:
When griffin submit multiple DQ jobs to livy, the http parameter `name` is always griffin. 
So livy will reject them.

job request :

```
[owner: null, request: [proxyUser: None, file: hdfs://nameservice-standby/user/kun.wan/measure-0.6.0-SNAPSHOT.jar, 
args: {
"spark" :

Unknown macro: { "log.level" }
,
"sinks" : [

Unknown macro: { .... }
],
"griffin.checkpoint" : [ ]
},{
"measure.type" : "griffin",
"id" : 5202,
"name" : "spu_null_check",
"owner" : "test",
"description" : "check null value for store and category",
"deleted" : false,
"timestamp" : 1568195100000,
"dq.type" : "PROFILING",
"sinks" : [ "ELASTICSEARCH", "HDFS" ],
"process.type" : "BATCH",
"rule.description" :

,
"data.sources" : [

Unknown macro: { .... }
],
"evaluate.rule" :

,
"measure.type" : "griffin"
},raw,raw, driverMemory: 1g, executorMemory: 6g, executorCores: 2, numExecutors: 6, queue: root.users.kun_dot_wan, name: griffin]]
```

livy Response : 

```
400 Bad Request
[Date:"Thu, 12 Sep 2019 10:00:00 GMT", Content-Type:"application/json;charset=utf-8", Content-Length:"47", Server:"Jetty(9.3.24.v20180605)"]
{"msg":"Duplicate session name: Some(griffin)"}
```

### DIFF
--- a/service/src/main/java/org/apache/griffin/core/job/SparkSubmitJob.java
+++ b/service/src/main/java/org/apache/griffin/core/job/SparkSubmitJob.java
@@ -204,7 +204,6 @@ public class SparkSubmitJob implements Job {
         args.add(finalMeasureJson);
         args.add("raw,raw");
         livyConfMap.put("args", args);
-        livyConfMap.remove("name");
     }
 
     protected void saveJobInstance(JobDetail jd) throws SchedulerException,
--- a/service/src/main/java/org/apache/griffin/core/job/SparkSubmitJob.java
+++ b/service/src/main/java/org/apache/griffin/core/job/SparkSubmitJob.java
@@ -204,6 +204,7 @@ public class SparkSubmitJob implements Job {
         args.add(finalMeasureJson);
         args.add("raw,raw");
         livyConfMap.put("args", args);
+        livyConfMap.remove("name");
     }
 
     protected void saveJobInstance(JobDetail jd) throws SchedulerException,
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/apache/griffin/core/job/SparkSubmitJob.java
```diff
@@ -204,6 +204,7 @@ private void setLivyArgs() throws IOException {
         args.add(finalMeasureJson);
         args.add("raw,raw");
         livyConfMap.put("args", args);
+        livyConfMap.remove("name");
```
guoyuepeng: I agree to remove duplicated names. 
But how should we handle the cases when users want to explicitly specify name by their job config?

wankunde: @guoyuepeng 
Instead of hard coding，we can explicitly specify the livy session name in the  configuration file. which is null by default.

Or we can use the DQ job name to be the livy session name?

guoyuepeng: Agree, 
we can use 
name_specified in config file | session_name

PR #535, opened 2019-09-12
[GRIFFIN-291]Relocate HttpClient code in measure jar using shade plugin

wankunde:
Now projects use different version of httpclient , and very easy to have conflicts with other components.

We can shade the httpclient sources into measure jar and the conflicts disappear.


guoyuepeng:
LGTM. 
Maven shade plugin is good for resolving version conflict issue.

wankunde:
@guoyuepeng I have added main-class into manifest.

guoyuepeng:
LGTM

PR #536, opened 2019-09-19
Fix metric output

iyuriysoft:
metric output from given parameter might be useful for building chain of queries

### DIFF
--- a/measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/DistinctnessExpr2DQSteps.scala
+++ b/measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/DistinctnessExpr2DQSteps.scala
@@ -217,7 +217,7 @@ case class DistinctnessExpr2DQSteps(context: DQContext,
       }
 
       // 8. distinct metric
-      val distTableName = "__distMetric"
+      val distTableName = ruleParam.getOutDfName()
       val distColName = details.getStringOrKey(_distinct)
       val distSql = {
         s"""
--- a/measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/UniquenessExpr2DQSteps.scala
+++ b/measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/UniquenessExpr2DQSteps.scala
@@ -145,7 +145,7 @@ case class UniquenessExpr2DQSteps(context: DQContext,
       uniqueRecordTransStep.parentSteps += groupTransStep
 
       // 7. unique metric
-      val uniqueTableName = "__uniqueMetric"
+      val uniqueTableName = ruleParam.getOutDfName()
       val uniqueColName = details.getStringOrKey(_unique)
       val uniqueSql = procType match {
         case BatchProcessType =>


bhlx3lyx7:
LGTM

wankunde:
Hi, @iyuriysoft @bhlx3lyx7 

In DistinctnessExpr2DQSteps, there are several `MetricWriteStep`:
```
      // 2. total metric
      val totalTableName = "__totalMetric"
      // 8. distinct metric
      val distTableName = "__distMetric"
          // 12. group by dup Record metric
          val groupDupMetricTableName = "__groupDupMetric"

          // 10. duplicate metric
          val dupMetricTableName = "__dupMetric"
```

In UniquenessExpr2DQSteps, there are several `MetricWriteStep`:
```
      // 5. total metric
      val totalTableName = "__totalMetric"

      // 7. unique metric
      val uniqueTableName = "__uniqueMetric"

        // 9. duplicate metric
        val dupMetricTableName = "__dupMetric"
```

If only parts of the metrics sinks with the `out.dataframe.name`, looks strange.

bhlx3lyx7:
@wankunde you are right, there are several `MetricWriteStep` in these two kinds of measures, all of them consist into the final metric like `{"total": 100, "unique": 10, "duplicate": 90}`. 
It is natural to export the unique metric as `out.dataframe.name` table, but I think all the output metrics are important for this measure.
Maybe it would be better to build a new table containing all the output metrics in different columns, and export as `out.dataframe.name`, then it would make sense right?
How do you think about it ? @wankunde @iyuriysoft

iyuriysoft:
Hi! yes, it makes sense, I'll fix it

wankunde:
In my option,  we can get the output metrics name from the rule configuration. For example, we can add parameter "distMetricName" or "uniqueMetricName",  then we can change the dataset name to  this value .if the value is null, use the default value instead. 
 
Rule json:
```json
{
  "evaluate.rule": {
    "rules": [
      {
        "dsl.type": "griffin-dsl",
        "dq.type": "distinct",
        "out.dataframe.name": "dist",
        "rule": "user_id",
        "details": {
          "source": "source",
          "target": "target",
          "total": "total",
          "distinct": "distinct",
          "dup": "dup",
          "num": "num",
          "duplication.array": "dup",
          "distMetricName": "xxxxx",
          "uniqueMetricName" : "yyyyyy"
        },
        "out":[
          {
            "type": "metric",
            "name": "distinct"
          }
        ]
      }
    ]
  }
}
```

Get the output metrics name: 
```scala
val _distMetricName = "distMetricName"
val _uniqueMetricName = "uniqueMetricName"
val details = ruleParam.getDetails

val distTableName = details.getString(_distMetricName, "__distMetric")
val uniqueTableName = details.getString(_uniqueMetricName, "__uniqueMetric")
```

The way  is more simple and  compatible with previous versions.

@iyuriysoft @bhlx3lyx7 How do you think about it ?

wankunde:
Hi, @iyuriysoft @bhlx3lyx7 

It's very import to customize metrics out step  which is helpful for user to get griffin's results.
Do you have any idea to update this PR?

guoyuepeng:
It is a pretty aged pr, I will close it temporarily,
But we still can reopen it when it is revised.

Thanks.

PR #541, opened 2019-09-30
GRIFFIN-293: [Service] livy.need.queue=true

neveljkovic:
This is how we fixed issue described in https://issues.apache.org/jira/browse/GRIFFIN-293
Solution is deployed to our servers and works OK.


wankunde:
LGTM

guoyuepeng:
OK, Thanks.

PR #543, opened 2019-10-10
fix tests failure

ahutsunshine:
1. fix hive connect failure
2.improve test run time

### DIFF
--- a/service/src/test/java/org/apache/griffin/core/job/EventServiceTest.java
+++ b/service/src/test/java/org/apache/griffin/core/job/EventServiceTest.java
@@ -34,6 +34,7 @@ import org.apache.griffin.core.job.entity.JobDataSegment;
 import org.apache.griffin.core.measure.entity.GriffinMeasure;
 import org.apache.griffin.core.measure.entity.Measure;
 import org.apache.griffin.core.util.EntityMocksHelper;
+import org.apache.hadoop.hive.metastore.IMetaStoreClient;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
@@ -41,6 +42,7 @@ import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest;
 import org.springframework.boot.test.autoconfigure.orm.jpa.TestEntityManager;
+import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Configuration;
@@ -58,6 +60,9 @@ public class EventServiceTest {
 
     @Autowired
     private List<GriffinEvent> eventList;
+
+    @MockBean
+    private IMetaStoreClient client;
 
     @Before
     public void setup() throws Exception {
--- a/service/src/test/java/org/apache/griffin/core/job/JobInstanceBeanRepoTest.java
+++ b/service/src/test/java/org/apache/griffin/core/job/JobInstanceBeanRepoTest.java
@@ -34,12 +34,14 @@ import org.apache.griffin.core.job.entity.JobInstanceBean;
 import org.apache.griffin.core.job.entity.LivySessionStates;
 import org.apache.griffin.core.job.entity.VirtualJob;
 import org.apache.griffin.core.job.repo.JobInstanceRepo;
+import org.apache.hadoop.hive.metastore.IMetaStoreClient;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest;
 import org.springframework.boot.test.autoconfigure.orm.jpa.TestEntityManager;
+import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.context.annotation.PropertySource;
 import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Pageable;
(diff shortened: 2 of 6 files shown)


service/src/test/java/org/apache/griffin/core/job/EventServiceTest.java
```diff
@@ -59,6 +61,9 @@ Licensed to the Apache Software Foundation (ASF) under one
     @Autowired
     private List<GriffinEvent> eventList;
 
+    @MockBean
```
wankunde: I don't know if the `IMetaStoreClient client` could be lazy initialized in `HiveMetaStoreProxy` ?
```
@Bean
    public IMetaStoreClient initHiveMetastoreClient() {
        HiveConf hiveConf = new HiveConf();
        hiveConf.set("hive.metastore.local", "false");
        hiveConf.setIntVar(HiveConf.ConfVars.METASTORETHRIFTCONNECTIONRETRIES,
            3);
        hiveConf.setVar(HiveConf.ConfVars.METASTOREURIS, uris);
        hiveConf.setIntVar(HiveConf.ConfVars.HMSHANDLERATTEMPTS, attempts);
        hiveConf.setVar(HiveConf.ConfVars.HMSHANDLERINTERVAL, interval);
        try {
            client = HiveMetaStoreClient.newSynchronizedClient(new HiveMetaStoreClient(hiveConf));
        } catch (Exception e) {
            LOGGER.error("Failed to connect hive metastore. {}", e);
        }
        return client;
    }
```
If not , `TestEntityManager` is been used in several test cases , Could we create a new  class to wrap it?

guoyuepeng:
@ahutsunshine 
test failed because docker cannot get up for bind address is already in use?

ahutsunshine:
> @ahutsunshine
> test failed because docker cannot get up for bind address is already in use?

Now it seems tests are passed.

service/src/test/java/org/apache/griffin/core/job/EventServiceTest.java
ahutsunshine: > I don't know if the `IMetaStoreClient client` could be lazy initialized in `HiveMetaStoreProxy` ?

I'm not sure whether it works even the `IMetaStoreClient client` could be lazy initialized. I guess it will not work as `IMetaStoreClient client` still need be injected in `HiveMetaStoreServiceImpl` when using @DataJaTest.

wankunde:
@guoyuepeng @ahutsunshine 
when griffin start a local spark context for unit test, but the jvm allocates memory failed.  
Could we increase the container memory ?

```
[1570713099788] unique_batch start: local-1570713077883
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 147324928 bytes for committing reserved memory.
# An error report file with more information is saved as:
Java HotSpot(TM) 64-Bit Server VM warning: INFO: os::commit_memory(0x0000000095380000, 147324928, 0) failed; error='Cannot allocate memory' (errno=12)
# /home/travis/build/apache/griffin/measure/hs_err_pid12006.log
```

guoyuepeng:
@wankunde 
I am not sure whether it increases pressure for our shared CI host.
How many memory do you want to use?
https://github.com/apache/griffin/blob/master/.travis.yml#L46

wankunde:
@guoyuepeng  Could we remove the jvm parameter `-Xms500m -Xmx1g`?
For example in  druid: https://github.com/alibaba/druid/blob/e91f383feda05f4e9dd2fefc2bdc6b3c08573b48/.travis.yml#L2

guoyuepeng:
@joohnnie  Could you check this ?

https://github.com/apache/griffin/commit/5179d61bd460d27ae0938ebd16b29cb6465d6bfe

Thanks,
William

joohnnie:
@guoyuepeng and @wankunde 
The VM option change is mainly for Service module and I think the memory is not enough
for Spark related application.

I think we can remove the following VM config for Maven and see if the issue can be fixed.
-Xms500m -Xmx1g -XX:MaxPermSize=256m  

Thanks
Johnnie

joohnnie:
Hi @guoyuepeng,

For the spark memory issue, another one might relate to this ticket

https://github.com/apache/griffin/pull/531

Before the ticket,  some spark related test cases use "SharedSparkContext" which means they share the same SparkContext. 

If the maven Surefire plugin for Measure module is running the spark testing parallelly, it might hit the memory issue due to multiple SparkContext objects are created.

Thanks
Johnnie

wankunde:
Hi @ahutsunshine 

I import `SparkSuiteBase` instead of spark-testing-base jar because of  spark-testing-base jar is not compatible with some spark versions.

SparkSuiteBase is mainly the same with SharedSparkContext.All the tests share the same SparkContext in one test suite.User can use SparkSuiteBase in the same way with SharedSparkContext.

https://github.com/apache/griffin/blob/e9dda8b53aadfc127854a89ac797c761b95de92a/measure/src/test/scala/org/apache/griffin/measure/SparkSuiteBase.scala#L28

https://github.com/apache/spark/blob/484f93e25506f84d1548504783be9ce940149bb7/core/src/test/scala/org/apache/spark/SharedSparkContext.scala#L24

But spark context may cache some RDD data in memory, so it needs more memory when running test case.The spark context will destroy and free the memory after all the tests is done in the test suite.
 
So I suggest If we can remove the following VM config for Maven and see if the issue can be fixed.
-Xms500m -Xmx1g -XX:MaxPermSize=256m

guoyuepeng:
OK, let's close this PR.
And we will recreate another PR for Maven config.

guoyuepeng:
after removed maven_opt as described.

https://travis-ci.org/apache/griffin/builds/601285195

@wankunde @joohnnie

joohnnie:
@guoyuepeng 
I tried to build with the following Maven Options and they are running successfully.

export MAVEN_OPTS="-Xms500m -Xmx1g -XX:MaxPermSize=512m 
https://travis-ci.org/joohnnie/griffin/builds/601486973

and 
export MAVEN_OPTS="-Xms256m -Xmx1g -XX:MaxPermSize=256m 

https://travis-ci.org/joohnnie/griffin/builds/601561953

Looks like there is memory limit for each travis CI JVM.

Thanks
Johnnie

guoyuepeng:
OK, Thanks.

PR #545, opened 2019-10-29
[Griffin 294] bugfix for completness enumeration wrong sql

LittleZhao:
if there is only 'hive_none' in enumeration values list, it will generate wrong sql. Update code to fix this bug

### DIFF
--- a/measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/CompletenessExpr2DQSteps.scala
+++ b/measure/src/main/scala/org/apache/griffin/measure/step/builder/dsl/transform/CompletenessExpr2DQSteps.scala
@@ -18,6 +18,8 @@ under the License.
 */
 package org.apache.griffin.measure.step.builder.dsl.transform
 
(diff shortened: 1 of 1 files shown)


LittleZhao:
here is the bug detail

if the error.confs looks like
`"error.confs": [
          {
            "values": [
              "hive_none"
            ], 
            "type": "enumeration", 
            "column.name": "abc"
          }
        ]`

the where clause is `(((`abc` IN ())) OR (`abc` IS NULL))`, it is a wrong where clause. It should be `(`abc` IS NULL)`

The pr fixes this bug.

guoyuepeng:
hi, could you provide the related test case?

LittleZhao:
> hi, could you provide the related test case?

I do not find the UT entrance for `CompletenessExpr2DQSteps.getDQSteps`, could you please tell me how to add ut for this part?

wankunde:
LGTM

guoyuepeng:
ok, we will create a test case for it soon.

PR #546, opened 2019-10-31
[GRIFFIN-295]Limit the memory used by test case

wankunde:
The container memory size is 3G in travis, but out test cases always uses more than 3G memory, so `Cannot allocate memory` will be thrown.

```
Java HotSpot(TM) 64-Bit Server VM warning: INFO: os::commit_memory(0x00000000fe980000, 23592960, 0) failed; error='Cannot allocate memory' (errno=12)
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 23592960 bytes for committing reserved memory.
# An error report file with more information is saved as:
# /home/travis/build/apache/griffin/measure/hs_err_pid11948.log
# [ timer expired, abort... ]
```

There are two kind of programs in our tests, the maven main program and the tests run by maven-surefire-plugin and scalatest-maven-plugin.
If the memory is unlimited, test cases will occupy as much memory as possible  especially spark jobs.

Spark jobs will not free the memory until a full GC occurs , even if we have stopped the spark context .so we need to limit the momery used by test cases.

We can limit the maven memory used by set export MAVEN_OPTS=" -Xmx1024m -XX:ReservedCodeCacheSize=128m" , and we can limit the memory used by spark job tests by configuring the maven-surefire-plugin and scalatest-maven-plugin.

For example:
Before we limit the memory used, maven program occupy 1.5G memory and spark job occupy 1.8G memory.

After we limit the memory used, maven program occupy 1G memory and spark job occupy 1G memory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,6 @@ before_script:
 # keep 30, need change according to ci logs.
 script:
   - export MAVEN_SKIP_RC=true
-  - export MAVEN_OPTS=" -Dorg.slf4j.simpleLogger.defaultLogLevel=warn -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn"
+  - export MAVEN_OPTS=" -Xmx1g -XX:ReservedCodeCacheSize=128m -Dorg.slf4j.simpleLogger.defaultLogLevel=warn -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn"
   - mvn -B clean test -Dlogging.level.org.springframework=WARN
 
--- a/measure/pom.xml
+++ b/measure/pom.xml
@@ -215,23 +215,13 @@ under the License.
                     <scalaVersion>${scala.version}</scalaVersion>
                 </configuration>
             </plugin>
-            <!-- enable scalatest -->
-            <plugin>
-                <groupId>org.scalatest</groupId>
-                <artifactId>scalatest-maven-plugin</artifactId>
-                <version>1.0</version>
-                <executions>
-                    <execution>
-                        <id>test</id>
-                        <goals>
-                            <goal>test</goal>
-                        </goals>
-                    </execution>
-                </executions>
-            </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.scalatest</groupId>
+                <artifactId>scalatest-maven-plugin</artifactId>
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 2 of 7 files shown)


wankunde:
@guoyuepeng @ahutsunshine @joohnnie Could you help me to review the codes ? many thanks.

wankunde:
The first commit is failed not because there was not enough memory, but timed out. That's strange. So print more test case log messages for debugging.

joohnnie:
HI @wankunde ,

First, you are right, there is a memory limit for the Travis container.
As i mentioned in this PR https://github.com/apache/griffin/pull/543.
I copied my testing result below

I tried to build with the following Maven Options and they are running successfully.

export MAVEN_OPTS="-Xms500m -Xmx1g -XX:MaxPermSize=512m
https://travis-ci.org/joohnnie/griffin/builds/601486973

and
export MAVEN_OPTS="-Xms256m -Xmx1g -XX:MaxPermSize=256m

https://travis-ci.org/joohnnie/griffin/builds/601561953

Looks like there is memory limit for each travis CI JVM.

This kind of constraint is related to the Travis CI container. 
IMO, I think it is a good practice to not put this kind of limitation to our code and POM configs if it can be solved by the travis config .travis.yml.

To put it another way, if in the future we use other CI tools to replace Travis and there is no this kind 
of memory limitation, we need to change the config one more time to make it work on the new CI tools.


Based on my experience, the Travis issues mainly related to the 2 constraints, 
- one is memory 
- the other one is No Log in 10 minutes/ or too much log to exceed the log size.

That's why we need to carefully define the log level and memory.

Thanks
Johnnie

wankunde:
Hi, @joohnnie 

First, as I described in this PR,  the maven memory can be limited used by setting MAVEN_OPTS, which is required. But we also need to limit the spark jobs memory because JVM stores many unused objects before full GC occurs. SPARK project also sets this configuration .

https://github.com/apache/spark/blob/14337f68e328c3faad81d3051a56be080a31509d/pom.xml#L2329

https://github.com/apache/spark/blob/14337f68e328c3faad81d3051a56be080a31509d/pom.xml#L2379

Second, there was a `No output has been received in the last 10m0s` error, and I don't know what's wrong with the travis build container, so printing more log message helps troubleshoot.


https://travis-ci.org/apache/griffin/builds/605545271?utm_medium=notification&utm_source=github_status

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
The build has been terminated
```

What do you think, @joohnnie

guoyuepeng:
ok, let's try to change the log level.

joohnnie:
Hi @wankunde 

for Maven Options, I agree with you.

As to "no output has been received in the last 10m" issue, it's really hard to figure out the root cause. Need to check when this issue happened and which test case doesn't have output.
Sometimes, it might due to the container running very slow and the test case is stuck. 

Since it is a shared environment and at the time point, the VM might be impacted by other's jobs. (just guess, not sure). 

I don't have a solution but just change the Log level and test.

Thanks
Johnnie

PR #549, opened 2019-11-08
manul trigger job not work befor the timer first fired

liusuquan:
when you manual trigger the new job which not fired yet , triggerTime.getTime() will return a null value and cause null pointer exception.


guoyuepeng:
Thanks for your work, could you also provide test case for it?

guoyuepeng:
It is a pretty aged pr, I will close it temporarily,
But we still can reopen it when it is revised or fully discussed.

Thanks.

PR #552, opened 2019-11-12
[GRIFFIN-299]Add oracle jdk8 support in travis build phase

wankunde:
As Ubuntu Xenial has become the default Travis CI build environment, we may build fail. 

The workaround is to either add `dist: trusty` to your .travis.yml file or use `openjdk8`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@
 # or implied. See the License for the specific language governing permissions and limitations under
 # the License.
 
+sudo: required
 language: java scala node_js
 dist: trusty
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@
 # or implied. See the License for the specific language governing permissions and limitations under
 # the License.
 
+dist: trusty
 
 language: java scala node_js
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,6 @@
 # or implied. See the License for the specific language governing permissions and limitations under
 # the License.
 
-dist: trusty
-
 language: java scala node_js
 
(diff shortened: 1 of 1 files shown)


wankunde:
@guoyuepeng @ahutsunshine 

It seems that `oraclejdk8` can no longer work with the license "BCL". Can we change it to `openjdk8`?

```
$ ~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"
Ignoring license option: BCL -- using GPLv2+CE by default
install-jdk.sh 2019-10-16
Expected feature release number in range of 9 to 14, but got: 8
The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .
```

wankunde:
Oddly, `oraclejdk8` works well with `incubator-pinot` and other projects, but not with `Griffin`

https://travis-ci.org/apache/incubator-pinot/jobs/611159063?utm_medium=notification&utm_source=github_status

guoyuepeng:
ok, let us try openjdk first.

guoyuepeng:
the title should be use openjdk, instead of oraclejdk.

wankunde:
> the title should be use openjdk, instead of oraclejdk.

Yes, at first,  i just want to change the build environment to `trusty` to support `oraclejdk8` , but `trusty` seems to have some problems with `npm`. so I change the `oraclejdk8` to `openjdk8`. The title is confused.

PR #555, opened 2019-11-13
[GRIFFIN-297] Allow support for additional file based data sources

chitralverma:

The PR extends the current support beyond just Avro and Text for various file based data sources (Parquet, ORC, etc). 

 - Allows users to specify additional file based data sources like Parquet, CSV, TSV, ORC etc.
 - Allows data to be read directly from stand-alone files as well as directories present in both local/ distributed file systems.
 - Allows users to specify schema directly through options (useful for CSV/ TSV types).

A sample config looks like,

```
{
  "name": "source",
  "baseline": true,
  "connectors": [
    {
      "type": "file",
      "version": "1.7",
      "config": {
        "format": "parquet",
        "options": { 
          "k1": "v1",
          "k2": "v2"
        },
        "paths": [
          "/home/chitral/path/to/source/",
          "/home/chitral/path/to/test.parquet"
        ]
      }
    }
  ]
}

```
No

Griffin test suite. Some additional unit test has also been added.

### DIFF
--- a/measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnector.scala
+++ b/measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnector.scala
@@ -39,14 +39,19 @@ import org.apache.griffin.measure.utils.ParamUtil._
  * Currently supported formats like Parquet, ORC, AVRO, Text and Delimited types like CSV, TSV etc.
  *
  * Supported Configurations:
- *  - format : [[String]] specifying the type of file source (parquet, orc, etc.). Default: parquet
+ *  - format : [[String]] specifying the type of file source (parquet, orc, etc.).
  *  - paths : [[Seq]] specifying the paths to be read
  *  - options : [[Map]] of format specific options
  *  - skipOnError : [[Boolean]] specifying where to continue execution if one or more paths are invalid.
  *  - schema : [[Seq]] of {colName, colType and isNullable} given as key value pairs. If provided, this can
  * help skip the schema inference step for some underlying data sources.
+ *
+ * Some defaults assumed by this connector (if not set) are as follows:
+ *  - `delimiter` is \t for TSV format,
+ *  - `schema` is None,
+ *  - `header` is false,
+ *  - `format` is parquet
  */
-
 case class FileBasedDataConnector(@transient sparkSession: SparkSession,
                                   dcParam: DataConnectorParam,
                                   timestampStorage: TimestampStorage)
@@ -95,6 +100,14 @@ case class FileBasedDataConnector(@transient sparkSession: SparkSession,
     })
   }
 
+  /**
+   * Ensures the presence of schema either via `header` or `schema` options.
+   *
+   *  - If both are present, the preference will be given to `schema`. First row will be omitted
+   * if `header` is set to true, else will be included.
+   *  - If `schema` is defined, it must be valid.
+   *  - If neither is set, a fatal exception is thrown.
+   */
   private def validateCSVOptions(): Unit = {
     if (options.contains(Header) && config.contains(Schema)) {
       griffinLogger.warn(s"Both $Options.$Header and $Schema were provided. Defaulting to provided $Schema")
@@ -146,6 +159,14 @@ object FileBasedDataConnector extends Loggable {
   private val DefaultFormat: String = SQLConf.DEFAULT_DATA_SOURCE_NAME.defaultValueString
   private val SupportedFormats: Seq[String] = Seq("parquet", "orc", "avro", "text", "csv", "tsv")
 
+  /**
+   * Validates the existence of paths in a given sequence.
+   * Set option `skipOnError` to true to avoid fatal errors if any erroneous paths are encountered.
+   *
+   * @param paths       given sequence of paths
+   * @param skipOnError flag to skip erroneous paths if any
+   * @return
+   */
   private def getValidPaths(paths: Seq[String], skipOnError: Boolean): Seq[String] = {
     val validPaths = paths.filter(path =>
       if (HdfsUtil.existPath(path)) true
@@ -162,7 +183,19 @@ object FileBasedDataConnector extends Loggable {
     validPaths
   }
 
+  /**
+   * Adds methods implicitly to [[DataFrameReader]]
+   *
+   * @param dfr an instance of [[DataFrameReader]]
+   */
   implicit class Implicits(dfr: DataFrameReader) {
+
+    /**
+     * Applies a schema to this [[DataFrameReader]] if any.
+     *
+     * @param schemaOpt an optional Schema
+     * @return
+     */
     def withSchemaIfAny(schemaOpt: Option[StructType]): DataFrameReader = {
       schemaOpt match {
         case Some(structType) => dfr.schema(structType)
--- a/measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnector.scala
+++ b/measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnector.scala
@@ -1,0 +1,162 @@
+/*
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+ */
+package org.apache.griffin.measure.datasource.connector.batch
+
+import scala.util.{Failure, Success, Try}
+
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.types.StructType
+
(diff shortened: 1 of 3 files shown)


chitralverma:
@wankunde Can you please review this. Thanks

measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnector.scala
```diff
@@ -0,0 +1,162 @@
...
+      }
+    }
+  }
+
+  def data(ms: Long): (Option[DataFrame], TimeRange) = {
+    val validPaths = getValidPaths(paths, skipErrorPaths)
+
+    val dfOpt = {
+      val dfOpt = Some(
+        sparkSession.read
```
guoyuepeng: LGTM

wankunde:
@chitralverma 

Thanks for contribution.

chitralverma:
> @chitralverma
> 
> Thanks for contribution.

Thanks, @wankunde can you merge this?

measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnector.scala
```diff
@@ -0,0 +1,162 @@
...
+      val colName = fieldMap(ColName).toLowerCase
+      val colType = fieldMap(ColType).toLowerCase
+      val isNullable = Try(fieldMap(IsNullable).toLowerCase.toBoolean).getOrElse(true)
+
+      currentStruct.add(colName, colType, isNullable)
+    })
+  }
+
+  private def validateCSVOptions(): Unit = {
+    if (options.contains(Header) && config.contains(Schema)) {
```
wankunde: `Header` should be `true`, not contains. We can check the `Schema` option and `schemaSeq.isEmpty`, if it fails ,check `Header`.

measure/src/test/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnectorTest.scala
```diff
@@ -0,0 +1,199 @@
...
+      val delimiter = if (f.matches("csv")) "," else if (f.matches("tsv")) "\t" else ""
+      val configs = Map(
+        "format" -> f,
+        "paths" -> Seq(
+          s"file://${getClass.getResource(s"/files/person_table.$f").getPath}"
+        ),
+        "options" -> Map(
+          "header" -> "true",
+          "inferSchema" -> "true",
+          "delimiter" -> delimiter
```
wankunde: Can we add those default options? If not specify those option, use the default.

chitralverma: Default values for test case only or in general?

measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnector.scala
chitralverma: @wankunde , setting `header` as `true` will result in using the first row as the column names which name cause unexpected failures as the rules will be configured for actual column names.

This is the reason, both `header` and `schema` can be specified, and if both are specified, `schema` is given preference.

measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnector.scala
```diff
@@ -0,0 +1,162 @@
...
+
+  var format: String = config.getString(Format, DefaultFormat).toLowerCase
+  val paths: Seq[String] = config.getStringArr(Paths, Nil)
+  val schemaSeq: Seq[Map[String, String]] = config.getAnyRef[Seq[Map[String, String]]](Schema, Nil)
+  val skipErrorPaths: Boolean = config.getBoolean(SkipErrorPaths, defValue = false)
+
+  assert(SupportedFormats.contains(format),
+    s"Invalid format '$format' specified. Must be one of ${SupportedFormats.mkString("['", "', '", "']")}")
+
+  if (format == "csv") validateCSVOptions()
```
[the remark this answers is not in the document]
chitralverma: `validateCSVOptions()` is used specially for CSV format as the schema may be set from either `header` or `schema`, where `schema` is given the preference. If the format is Parquet and ORC, `schema` option will still work due to,

```
 .schema(currentSchema)
 .load(validPaths: _*)
```

chitralverma:
> @chitralverma
> 
> Thanks for contribution.
> 
> @guoyuepeng @chitralverma
> The csvReader in spark has changed since 2.4 version, but griffin still uses version 2.2.1 , I think we need more test cases for achieve compatibility.
> 
> https://issues.apache.org/jira/browse/SPARK-25669

Let me add more test cases for this

measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnector.scala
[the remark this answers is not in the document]
wankunde: The schema will aways be null for orc and parquet files, although we have provided the  schema now.

measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnector.scala
wankunde: if Header is set to false, `options.contains(Header)` will return true.

measure/src/test/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnectorTest.scala
wankunde: For tsv files, we should given a default delimiter, because we think of it as csv files.
I suggest that you add those default values in general, and user can change it by configuration.

measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnector.scala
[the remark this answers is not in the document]
chitralverma: Let me make the changes

chitralverma:
> > @chitralverma
> > Thanks for contribution.
> > @guoyuepeng @chitralverma
> > The csvReader in spark has changed since 2.4 version, but griffin still uses version 2.2.1 , I think we need more test cases for achieve compatibility.
> > https://issues.apache.org/jira/browse/SPARK-25669
> 
> Let me add more test cases for this

@wankunde I tried recreating the `SPARK-25669` bug with version 2.2.1 used by griffin but it seems like the issue exists no more. I can see in the jira that it was fixed in versions 2.3+, but right now its not appearing in version 2.2.1 

I've incorporated rest of your review comments. Please review at your convenience. 

Thanks

measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnector.scala
chitralverma: As per the current  implementation, in case if both `schema` and `header` are provided, the schema will be used but the first row will still be skipped.

measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnector.scala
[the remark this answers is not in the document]
chitralverma: Schema is now applied for all types if given

measure/src/test/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnectorTest.scala
chitralverma: Added following defaults:
 - default delimiter for TSV format as `\t`, 
 - default `schema` is `None`, 
 - `header` is assumed false if not set,
 - default `format` is `parquet`

measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/FileBasedDataConnector.scala
```diff
@@ -0,0 +1,207 @@
...
+    case _ => None
+  }
+
+  assert(SupportedFormats.contains(format),
+    s"Invalid format '$format' specified. Must be one of ${SupportedFormats.mkString("['", "', '", "']")}")
+
+  if (format == "csv") validateCSVOptions()
+  if (format == "tsv") {
+    format = "csv"
+    options = options + (Delimiter -> "\t")
```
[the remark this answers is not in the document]
chitralverma: Initial thought : 
Since TSV is a special case of CSV with delimiter as '\t', the delimter has been forcefully coded as tab ('\t'). If the delimter is anything else its no longer a TSV file and delimter can be provided with format as CSV.

I'll make the change anyways.

wankunde:
> > > @chitralverma
> > > Thanks for contribution.
> > > @guoyuepeng @chitralverma
> > > The csvReader in spark has changed since 2.4 version, but griffin still uses version 2.2.1 , I think we need more test cases for achieve compatibility.
> > > https://issues.apache.org/jira/browse/SPARK-25669
> > 
> > 
> > Let me add more test cases for this
> 
> @wankunde I tried recreating the `SPARK-25669` bug with version 2.2.1 used by griffin but it seems like the issue exists no more. I can see in the jira that it was fixed in versions 2.3+, but right now its not appearing in version 2.2.1
> 
> I've incorporated rest of your review comments alongwith some additional code documentation to ease the use by other developers. Please review at your convenience.
> 
> Thanks

Yes, that PR only affect `DataFrameReader.csv()` method while we use `DataFrameReader.load()` method.

wankunde:
LGTM

chitralverma:
@wankunde Great, can you merge this now. Thanks.

guoyuepeng:
LGTM, I will merge it.

PR #556, opened 2019-11-25
[GRIFFIN-301] Update custom data connector to have the same parameters as build-in data connector

wankunde:
Now custom data connectors have different parameters with build-in data connector, which will confuse the user.

For example : https://issues.apache.org/jira/browse/GRIFFIN-300


wankunde:
I will merge this PR if there is no other ideas.

PR #540, opened 2019-09-26
To support postgresql db as data source

MATRIX4284:
Added pom changes for supporting postgresql 11

PR #561, opened 2019-12-27
[Griffin-315] Adding JDBC based data connector

tusharpatil20:

JDBC based data connector to read data from different JDBC based data sources.

No

Griffin test suite.

### DIFF
--- a/measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/JDBCBasedDataConnector.scala
+++ b/measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/JDBCBasedDataConnector.scala
@@ -1,0 +1,129 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.griffin.measure.datasource.connector.batch
+
+import org.apache.spark.sql.{DataFrame, SparkSession}
+
+import org.apache.griffin.measure.Loggable
+import org.apache.griffin.measure.configuration.dqdefinition.DataConnectorParam
+import org.apache.griffin.measure.context.TimeRange
+import org.apache.griffin.measure.datasource.TimestampStorage
+import org.apache.griffin.measure.utils.ParamUtil._
+
+/**
+ * A batch data connector for file based sources which allows support various
+ * file based data sources like Parquet, CSV, TSV, ORC etc.
+ * Local files can also be read by prepending `file://` namespace.
+ *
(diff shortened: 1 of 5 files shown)


tusharpatil20:
Hi @wankunde 
Can you please review this PR ?

measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/JDBCBasedDataConnector.scala
```diff
@@ -0,0 +1,130 @@
...
+import org.apache.spark.sql.{DataFrame, SparkSession}
+
+import org.apache.griffin.measure.Loggable
+import org.apache.griffin.measure.configuration.dqdefinition.DataConnectorParam
+import org.apache.griffin.measure.context.TimeRange
+import org.apache.griffin.measure.datasource.TimestampStorage
+import org.apache.griffin.measure.utils.ParamUtil._
+
+/**
+ * A batch data connector for file based sources which allows support various
```
wankunde: This comment should be updated。

tusharpatil20: Hi @wankunde 
Please find my inline comments,

- This pr is similar to
measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/MySqlDataConnector.scala, so could you change MySqlDataConnector to a subclass of JDBCBasedDataConnector.scala. 
**If a user wants to use MySql based data source to read data then there is no direct mapping to use it. The user has to use this connector using CustomDataconnector. Using JDBCBaedData connector users can use any RDBMS which has a JDBC connector by providing driver class, adding a jar to classpath and connection details. Added default config for MySql database.** 
- Instead of adding a new dependency jar, i suggest using profile to add some other database support, like MariaDB and postgresql
**With this implementation Griffin can read data from any source which has JDBC connector by adding JDBC connector JAR to the classpath and providing connection details in a config file. I have added H2 database dependency to check the JDBC connector functionality**

tusharpatil20: Hi @wankunde 
Made the changes as per review comments in 425a21f, please review.

measure/src/main/scala/org/apache/griffin/measure/datasource/connector/batch/JDBCBasedDataConnector.scala
```diff
@@ -0,0 +1,127 @@
...
+      prop.setProperty("user", user)
+      prop.setProperty("password", password)
+      prop.setProperty("driver", driver)
+      val df: DataFrame = sparkSession.read.jdbc(url, s"($dtSql) as t", prop)
+      val dfOpt = Some(df)
+      val preDfOpt = preProcess(dfOpt, ms)
+      preDfOpt
+    } catch {
+      case e: Throwable =>
+        error(s"load mysql table $fullTableName fails: ${e.getMessage}", e)
```
wankunde: `mysql` keyword should not be included in error messages.

tusharpatil20: Hi @wankunde,
Made the changes as per review comments in 2ee6869500b813a2a95f56e5ba3b242cea711ffd, please review it.

wankunde:
LGTM

wankunde:
This closes #540

AurelCephas:
Hi. I'm new to Apache Griffin. I've installed it by following the docker compose tutorial. But when I'm trying to connect Griffin to my remote (even to the local) postgresql database I'm facing to an issue. It seems that the connection has not been etablished. Anyone can help me to slove this issue? Do I need to configure it manually in the scala code? And rebuild? Is it a matter of version? I've read on the Apache Griffin site web that the latest release has some bug about JDBC connection.
MicrosoftTeams-image (5)

PR #566, opened 2020-01-20
optimize get metric maps in 'MetricWriteStep'

XiaoyuBD:
**Why/What changes？**
In 'MetricWriteStep.getMetricMaps()' the dataframe was transformed to json rdd, and then collect, and then transformed to Seq[Map]. 
It's not elegant and hard to understand. More optimized way is to collect it first, and then transform it to Seq[Map] directly. 

We have test it with our DQ cases. It works well.


wankunde:
LGTM

wankunde:
@XiaoyuBD Thanks for your contribution. Any test case will be better.

PR #568, opened 2020-02-07
[GRIFFIN-323] Refactor configuration Data Source Connector

chitralverma:

This ticket proposes the following changes,

- remove 'version' from 'DataConnectorParam' as it is not being used anywhere in the codebase.
- change 'connectors' from array type to a single JSON object. Since a data source named X may only be of one type (hive, file etc), the connector field should not be an array.
- rename connectors to connector
- update existing config files and documentation for reference

Yes. As mentioned above, the config structure has changed now.

Griffin test suite.

### DIFF
--- a/griffin-doc/measure/measure-batch-sample.md
+++ b/griffin-doc/measure/measure-batch-sample.md
@@ -24,36 +24,30 @@ Apache Griffin measures consist of batch measure and streaming measure, this doc
 ```
 {
   "name": "accu_batch",
-
   "process.type": "BATCH",
-
   "data.sources": [
     {
       "name": "source",
       "baseline": true,
-      "connectors": [
-        {
-          "type": "AVRO",
-          "version": "1.7",
-          "config": {
-            "file.name": "src/test/resources/users_info_src.avro"
-          }
+      "connector": {
+        "type": "AVRO",
+        "version": "1.7",
+        "config": {
+          "file.name": "src/test/resources/users_info_src.avro"
         }
-      ]
-    }, {
+      }
+    },
+    {
       "name": "target",
-      "connectors": [
-        {
-          "type": "AVRO",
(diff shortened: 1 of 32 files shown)


wankunde:
LGTM

pshingavi:
I do not see a change in the service code to accept this change in contract. When I submit with 'connectors' from the UI the request goes through but 'connector' does not work in the service API call to submit a measure. Am I missing something here?

chitralverma:
Yes these changes are not supported by the current ui and services

pshingavi:
What's the compatible commit I can use in this case ? Just before this was merged may be ?

Rayleigh0727:
> I do not see a change in the service code to accept this change in contract. When I submit with 'connectors' from the UI the request goes through but fails while executing the job (log snippet below) and with 'connector' does not work in the service API call since it's expecting 'connectors' to submit a measure. Am I missing something here?
> `20/03/19 18:27:26 ERROR Application$: assertion failed: Connector is undefined or invalid java.lang.AssertionError: assertion failed: Connector is undefined or invalid at scala.Predef$.assert(Predef.scala:170) at org.apache.griffin.measure.configuration.dqdefinition.DataSourceParam.validate(DQConfig.scala:100)`

hi,I encountered the same problem, did you solve it?

FaizanSh:
> Yes these changes are not supported by the current ui and services

Hi, this problem still presist in code. What workaround do you suggest?

chitralverma:
Hi Everyone, Please note that these changes are done in the `measure` module only. 

To get a more stable release, use the version 0.5.0 available in the maven central.

Otherwise, fix PRs for `services` and `ui` module are most welcome.

zgdong1987:
I changed to the version 0.5.0, but failed to compile the UI

PR #570, opened 2020-04-21
fix bug - when OutDfName is null

gaodb1210:
fix bug - when OutDfName is null

if  ruleParam.getOutDfName() is null, the param "name" used by SparkSqlTransformStep will be different from used by DirectWriteStep, which will cause execute error : can not find table or view.....


chitralverma:
@Mr-Gdb is this PR still active?

Mr-Gdb:
> @Mr-Gdb is this PR still active?

yes

PR #574, opened 2020-06-12
Fix Unit Test Issue In Measure Test Component

toyboxman:
[GRIFFIN-329] Measure unit test cases failed on the condition of no docker image

The unit test case tries to download a ES docker image and run the following cases. if the downloading fails, some cases will abort due to exceptions. In the revision, a new flag is introduced in execution, unless the docker image is avaiable, some cases will be excluded.


toyboxman:
@guoyuepeng 

look like the default CI is not available, do you know restore or pass it

PR #582, opened 2020-09-03
[griffin-service] use service tar.gz deploy and griffin.sh start

ambition119:
If we use` java -jar` start service ,then It is inconvenient to modify the configuration file. Modifying the configuration file requires recompiling the jar. And it is inconvenient to stop the service.
   This PR provides `tar.gz` installation method and shell startup method，and it has been deployed and run in our production environment.

1. ls -al service-0.6.0-SNAPSHOT
     bin
     config
     lib

2. cd service-0.6.0-SNAPSHOT/

3. ./bin/griffin.sh start

4. jps
   17860 GriffinWebApplication

5. ./bin/griffin.sh stop
    stopping 17860 of service ...
 
  If start service, we can access http://${ip}:8080
image

### DIFF
--- a/griffin-doc/deploy/deploy-guide.md
+++ b/griffin-doc/deploy/deploy-guide.md
@@ -533,21 +533,27 @@ Similar changes are required in `env_streaming.json`.
 }
 ```
 
-It's easy to build Griffin, just run maven command `mvn clean install`. Successfully building, you can get two jars `service-0.4.0.jar`,`measure-0.4.0.jar` from target folder in service and measure module.
+It's easy to build Griffin, just run maven command `mvn clean install`. Successfully building, you can get `service-${version}.tar.gz` and `measure-${version}.jar` from target folder in service and measure module.
 
 Upload measure's jar to hadoop folder.
 ```
 # change jar name
 mv measure-0.4.0.jar griffin-measure.jar
-mv service-0.4.0.jar griffin-service.jar
 # upload measure jar file
 hdfs dfs -put griffin-measure.jar /griffin/
 ```
 
-Startup service.jar，run Griffin management service.
+Startup service，run Griffin management service.
 ```
-cd $GRIFFIN_HOME
-nohup java -jar griffin-service.jar>service.out 2>&1 &
+cd $GRIFFIN_INSTALL_DIR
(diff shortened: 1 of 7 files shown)


ambition119:
> Thanks for your PR.
> 
> It is weird that our ci failed.

I view ci log content, like error :
```log
2020-09-03 13:40:31.648 ERROR 11355 --- [           main] o.a.g.c.m.h.HiveMetaStoreService        [79]  : Can not get databases : {}
org.apache.hadoop.hive.metastore.api.MetaException: null
2020-09-03 13:40:31.657 ERROR 11355 --- [           main] o.a.g.c.m.h.HiveMetaStoreService        [79]  : Can not get databases : {}
2020-09-03 13:40:37.639 ERROR 11355 --- [           main] o.a.g.c.u.PropertiesUtil                [49]  : Get properties from .././quartz.properties failed. java.io.FileNotFoundException: class path resource [../quartz.properties] cannot be opened because it does not exist
2020-09-03 13:40:45.645 ERROR 11355 --- [           main] o.a.g.c.u.FSUtil                        [98]  : Can not get hdfs file system. {}
java.lang.IllegalStateException: Variable substitution depth too large: 20 ${fs.defaultFS}
2020-09-03 13:40:45.679 ERROR 11355 --- [           main] o.a.g.c.j.SparkSubmitJob                [110]  : Post spark task ERROR.
java.lang.ClassCastException: org.quartz.impl.JobDetailImpl cannot be cast to org.quartz.JobExecutionContext
2020-09-03 13:40:47.895 ERROR 11355 --- [ool-11-thread-1] o.s.s.s.TaskUtils$LoggingErrorHandler   [96]  : Unexpected error occurred in scheduled task.
java.lang.IllegalArgumentException: Cannot find cache named 'hive' for Builder[public void org.apache.griffin.core.metastore.hive.HiveMetaStoreServiceImpl.evictHiveCache()] caches=[hive] | key='' | keyGenerator='cacheKeyGenerator' | cacheManager='' | cacheResolver='' | condition='',true,true
```
    
But I did not find the content related to the build script of ci. It should be normal after the script `jar `is replaced with `tar.gz`. do you have any good advice?

guoyuepeng:
LGTM

PR #588, opened 2020-12-02
Fix doc format glitches

toyboxman:


guoyuepeng:
Thanks for fixing it!

PR #592, opened 2021-06-15
For task GRIFFIN-363:

amazingSaltFish:
url:https://issues.apache.org/jira/projects/GRIFFIN/issues/GRIFFIN-363?filter=allopenissues
1. add new property hive.krb5conf.path.
2. solve hive metastore kerberos error.

### DIFF
--- a/service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreProxy.java
+++ b/service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreProxy.java
@@ -69,11 +69,11 @@ public class HiveMetaStoreProxy {
     private String keytabUser;
 
     @Value("${hive.need.kerberos}")
-    private Boolean needKerberos;
+    private String needKerberos;
 
     @PostConstruct
     public void init() throws IOException {
-        if (needKerberos && hiveKrb5confPath != null) {
+        if ( needKerberos != null && "true".equalsIgnoreCase(needKerberos) && hiveKrb5confPath != null) {
             System.setProperty("java.security.krb5.conf", hiveKrb5confPath);
             UserGroupInformation.loginUserFromKeytab(keytabUser, keytabPath);
         }else {
--- a/service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreProxy.java
+++ b/service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreProxy.java
@@ -19,16 +19,20 @@ under the License.
 
 package org.apache.griffin.core.metastore.hive;
 
+import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
 
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.metastore.HiveMetaStoreClient;
 import org.apache.hadoop.hive.metastore.IMetaStoreClient;
+import org.apache.hadoop.security.UserGroupInformation;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.stereotype.Component;
+
+import java.io.IOException;
 
 @Component
 public class HiveMetaStoreProxy {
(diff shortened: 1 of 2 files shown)


lovelyqincai:
why my pr has fail,who can tell me what should i do?

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreProxy.java
```diff
@@ -55,6 +59,28 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+    private String hiveKrb5confPath;
+
+    @Value("${hive.keytab.path}")
+    private String keytabPath;
+
+    @Value("${hive.keytab.user}")
+    private String keytabUser;
+
+    @Value("${hive.need.kerberos}")
+    private String needKerberos;
```
chitralverma, line 72: Can be a boolean.

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreProxy.java
```diff
@@ -55,6 +59,28 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+
+    @PostConstruct
+    public void init() throws IOException {
+        if ( needKerberos != null && "true".equalsIgnoreCase(needKerberos) && hiveKrb5confPath != null) {
+            System.setProperty("java.security.krb5.conf", hiveKrb5confPath);
```
chitralverma, line 77: This can be a constant.

chitralverma (changes requested):

chitralverma:
@amazingSaltFish can you add some test cases and add  more details to this JIRA and this PR.

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreProxy.java
lovelyqincai: emmm,last pr for this variable 'hive.need.kerberos' i also think it type is boolean.
but in  org.apache.griffin.core.metastore.hive.HiveMetaStoreServiceJdbcImpl line 73: needKerberos define as string.

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreProxy.java
```diff
@@ -55,6 +59,28 @@ Licensed to the Apache Software Foundation (ASF) under one
 
     private IMetaStoreClient client = null;
 
+    @Value("${hive.krb5conf.path}")
+    private String hiveKrb5confPath;
+
+    @Value("${hive.keytab.path}")
```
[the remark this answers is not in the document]
lovelyqincai: ok, I accept your suggestion

lovelyqincai:
> @amazingSaltFish can you add some test cases and add more details to this JIRA and this PR.

I lost this scene, but I solved the problem this way...

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreProxy.java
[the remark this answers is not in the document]
lovelyqincai: > Since kerberos doesnt apply to just hive but hadoop services in general, the "hive." prefix can be removed from all these configs.
emmm, kerberos have many keytabs for different user,like hive,hdfs,livy and so on.
i can see many keytab files name like hive.keytab,hdfs.keytab...
how about keep the prefix?

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreProxy.java
wankunde: Agree with @chitralverma 

@lovelyqincai Could we  change `needKerberos` variable in `org.apache.griffin.core.metastore.hive.HiveMetaStoreServiceJdbcImpl` to boolean type in another PR.

service/src/main/resources/application.properties
```diff
@@ -22,6 +22,9 @@ spring.datasource.password=123456
 spring.jpa.generate-ddl=true
 spring.datasource.driver-class-name=org.postgresql.Driver
 spring.jpa.show-sql=true
+# kerberos
+# add new configuration for kerberos file
+krb5conf.path=/path/to/krb5conf/file
```
wankunde, line 27: Should we comment this configuration by default ?

service/src/main/java/org/apache/griffin/core/metastore/hive/HiveMetaStoreProxy.java
lovelyqincai: ok,i also agree with.

lovelyqincai: @wankunde 

> Agree with @chitralverma
> 
> @lovelyqincai Could we change `needKerberos` variable in `org.apache.griffin.core.metastore.hive.HiveMetaStoreServiceJdbcImpl` to boolean type in another PR.

ok,i also agree with.

service/src/main/resources/application.properties
lovelyqincai: Excuse me,I have a little confused,would you explain it for me?

lovelyqincai: > Should we comment this configuration by default ?
Excuse me,I have a little confused,would you explain it for me?

PR #596, opened 2022-01-17
[GRIFFIN-367] For task GRIFFIN-367, update local deploy document.

lipzhu:
1. Remove the diff change history for hive-site.xml.
2. Update the deployment doc for service module package location.

No.

Unit Tests

### DIFF
--- a/service/pom.xml
+++ b/service/pom.xml
@@ -385,6 +385,7 @@ under the License.
                 <configuration>
                     <appendAssemblyId>false</appendAssemblyId>
                     <skipAssembly>false</skipAssembly>
+                    <outputDirectory>../target</outputDirectory>
                     <descriptors>
                         <descriptor>src/main/resources/assembly/assembly.xml</descriptor>
                     </descriptors>
--- a/service/pom.xml
+++ b/service/pom.xml
@@ -385,7 +385,6 @@ under the License.
                 <configuration>
                     <appendAssemblyId>false</appendAssemblyId>
                     <skipAssembly>false</skipAssembly>
-                    <outputDirectory>../target</outputDirectory>
                     <descriptors>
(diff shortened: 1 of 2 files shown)


service/pom.xml
```diff
@@ -385,7 +385,6 @@ under the License.
                 <configuration>
                     <appendAssemblyId>false</appendAssemblyId>
                     <skipAssembly>false</skipAssembly>
-                    <outputDirectory>../target</outputDirectory>
```
whhe, line 388: Why move it?  I think it's appropriate to put the deployment package at root directory.

lipzhu: The service-${version}.tar.gz location is not align with the Documention in https://github.com/apache/griffin/blob/master/griffin-doc/deploy/deploy-guide.md, or we can update the documentation to remove the confuse.  

>>>It's easy to build Griffin, just run maven command mvn clean install. Successfully building, you can get service-${version}.tar.gz and measure-${version}.jar from target folder in **service** and **measure** module.

whhe: > The service-${version}.tar.gz location is not align with the Documention in https://github.com/apache/griffin/blob/master/griffin-doc/deploy/deploy-guide.md, or we can update the documentation to remove the confuse.
> 
> > > > It's easy to build Griffin, just run maven command mvn clean install. Successfully building, you can get service-${version}.tar.gz and measure-${version}.jar from target folder in **service** and **measure** module.

You're right. I think it's better to update the doc, and the following commands should also be modified at the same time.

```bash
cd $GRIFFIN_INSTALL_DIR
tar -zxvf service-${version}.tar.gz
```

chitralverma: @lipzhu please update the docs instead

chitralverma (changes requested):

whhe (approved):

PR #604, opened 2022-07-08
Resolving type compatibility conflicts in ElasticSearchGriffinDataConnector

complone:
Reason:

env: mac Big Sur m1 11.2.3

type1

During the dataBySql conversion process, because the load parameter of httpUrl does not conform to the construction specification of parallelize when it is passed in before the RDD construction, it is fixed.

test result:
4211657331715_ pic
4221657331715_ pic


complone:
By observing the CI log, I found that the hive metastore did not start automatically, it was not caused by this pull request

PR #613, opened 2022-10-27
Griffin 1.0.0 dev

wangding75:


core/pom.xml
```diff
@@ -297,6 +297,14 @@ under the License.
             <artifactId>elasticsearch-rest-client</artifactId>
             <version>${elasticsearch-rest-client.version}</version>
         </dependency>
+
+        <!--   新增依赖     -->
```
guoyuepeng, line 301: ding, could you translate all comments into english?

PR #628, opened 2023-02-18
JDBC Connector SQL mode

XiaoyuBD:
# What changes were proposed in this pull request?
JDBC connector could configuring sql which would get sql results as datasource.
It is more flexible to use. And in some special cases (such as only a small number of columns need to be extracted), it can reduce the IO of loading data from db.

# Does this PR introduce any user-facing change?
No. The original configuration mode is not affected.

# How was this patch tested?
Unit Tests. And it has been used in our production environment for several months.

### DIFF
--- a/griffin-doc/measure/measure-configuration-guide.md
+++ b/griffin-doc/measure/measure-configuration-guide.md
@@ -547,6 +547,7 @@ List of supported data connectors:
 | Name       | Type     | Description                            | Default Values |
 |:-----------|:---------|:---------------------------------------|:-------------- |
 | database   | `String` | database name                          | default |
+| sql        | `String` | query sql                              | `Empty` |
 | tablename  | `String` | table name to be read                  | `Empty` |
 | url        | `String` | the connection string URL to database  | `Empty` |
 | user       | `String` | user for connection to database        | `Empty` |
@@ -554,6 +555,7 @@ List of supported data connectors:
 | driver     | `String` | driver class for JDBC connection to database | com.mysql.jdbc.Driver |
 | where      | `String` | condition for reading data from table  | `Empty` |
 
+- The config `sql` and `tablename` cannot be all empty. If sql is not empty, connector will extract the result of sql.
 - Example:
    ```
   "connector": {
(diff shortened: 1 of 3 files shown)


XiaoyuBD:
@guoyuepeng Can you review this PR? Thanks.

griffin-doc/measure/measure-configuration-guide.md
```diff
@@ -547,13 +547,15 @@ List of supported data connectors:
 | Name       | Type     | Description                            | Default Values |
 |:-----------|:---------|:---------------------------------------|:-------------- |
 | database   | `String` | database name                          | default |
+| sql        | `String` | query sql                              | `Empty` |
```
chitralverma, line 550: is `sql` mode mutually exclusive to the usual table db mode or does it work with them? this needs to be documented somewhere.

Also how does this work now with the `where` functionality ?

chitralverma:
@XiaoyuBD  thanks for the PR.
I have some questions about the purpose of the changes. 

If you dont want to query unwanted columns or even transform the data before applying checks, you can put 1 or more queries in the `pre.proc` section of any data source. This has a far bigger scope that just JDBC as it applies to all types of sources.

This will also support projection and predicate push down to the source which will allow much less IO.

XiaoyuBD:
> purpose

You're right, using **pre.proc** can also achieve the effect I described earlier. My previous description gave an inappropriate example. In fact, the main benefit of SQL mode is that it is very flexible. For data developers, they can express various datasets in SQL, including join multiple tables to extract and analyze query results.

griffin-doc/measure/measure-configuration-guide.md
XiaoyuBD: Thanks for the comment.  If the config `sql` was provided, `database`, `tablename` and `where` can all be ignored, connector will extract the result of sql. I added that description to the MD and added the SQL mode config example.

guoyuepeng:
LGTM.

@chitralverma any comments?

XiaoyuBD:
@chitralverma Hi, is there any other comments? This PR has been open for a long time.

chitralverma:
LGTM, merging.

whhe:
> LGTM, @guoyuepeng can you please merge thi, for some reason the merge pr script is not working on my system

There is no related Jira ticket, I think it's OK to use `squash and merge` on GitHub.

Besides, the ASF projects migrated to GitHub a long time ago, I think we can move all stuffs (like Jira and Wiki) to GitHub now. @guoyuepeng @chitralverma Maybe we can discuss it elsewhere.

PR #659, opened 2024-07-24
Update griffin submodules pom structure

toyboxman:
1.Set submodule group-id as same as parent's.


toyboxman:
@whhe , Could you review this change.

whhe (approved):
LGTM

PR #661, opened 2024-08-13
Griffin 2.0.0 dev

guoyuepeng:
basic metric

### DIFF
--- a/griffin-metric/src/main/java/org/apache/griffin/metric/model/MetricD.java
+++ b/griffin-metric/src/main/java/org/apache/griffin/metric/model/MetricD.java
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!--
+/*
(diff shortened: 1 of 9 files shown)


griffin-metric/src/main/java/org/apache/griffin/metric/model/MetricD.java
```diff
@@ -0,0 +1,42 @@
...
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+*/
+package org.apache.griffin.metric.model;
+
+import lombok.*;
```
whhe, line 21: We should not use import with `*`, you'd better add a code style check pipeline based on GitHub Action.

PR #662, opened 2024-08-18
Initialize base metric model

toyboxman:
Propose a basic and elementary metric model definition.

### DIFF
--- a/griffin-connectors/pom.xml
+++ b/griffin-connectors/pom.xml
@@ -28,6 +28,7 @@ under the License.
         <version>2.0.0-SNAPSHOT</version>
     </parent>
 
+
     <artifactId>griffin-connectors</artifactId>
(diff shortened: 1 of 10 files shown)


toyboxman:
@guoyuepeng , Can you review my update on your change?

guoyuepeng:
> @guoyuepeng , Can you review my update on your change?

when users want to publish metric value, we might also attach tags in the coming metric value?
should we update this part in metricv?

toyboxman:
> > @guoyuepeng , Can you review my update on your change?
> 
> when users want to publish metric value, we might also attach tags in the coming metric value? should we update this part in metricv?

Good point.
@guoyuepeng I have added tag property into metricV entity, and complemented java doc.

guoyuepeng:
LGTM, merged to griffin-2.0.0-dev.

PR #665, opened 2024-09-18
add service interface for metricD table.

toyboxman:


guoyuepeng:
cool, maybe we can rename this module as some kind of DAO module, and separate common model into common module.

PR #673, opened 2024-12-23
refresh pom structure and enable unit test

toyboxman:


griffin-bom/pom.xml
```diff
@@ -28,16 +28,17 @@ under the License.
         <version>2.0.0-SNAPSHOT</version>
     </parent>
 
-    <groupId>org.apache.griffin</groupId>
```
guoyuepeng, line 31: The griffin-bom module is used to manage the version of third-party dependencies. If users want to import griffin-metric to your project, they need to import griffin-bom together by import-bom way, this can help users to manage the version.

but for test or dev purposed dependencies, they should not declared in bom, that means our users will import test dependencies to use griffin, that is not necessary.

these dev or test dependencies, should be managed by griffin-parent in root pom, so all modules can use the same dev env.

dev and test are not user oriented artifacts.

griffin-doc/dev-2/new-metric-build.md
```diff
@@ -98,6 +98,7 @@ T_TAG_D
       
         Response:
         {
+           "id": 2,
```
guoyuepeng, line 101: every metric will have its id, that is right.

griffin-metric/src/main/java/org/apache/griffin/metric/entity/MetricV.java
```diff
@@ -44,14 +46,22 @@ Licensed to the Apache Software Foundation (ASF) under one
 public class MetricV extends BaseEntity {
 
     /**
-     * An unique identity for a metric.
+     * An incremental identity in metric value table.
      */
-    @TableId(value="mid", type = IdType.AUTO)
+    @TableId(value="vid", type = IdType.AUTO)
```
guoyuepeng, line 51: this looks good.

griffin-bom/pom.xml
toyboxman: you mean all dependecies about dev/test purpose should be removed from the pom under bom project. Am I right?

guoyuepeng: yes, move test/dev dependencies to root(parent) pom.

toyboxman: @guoyuepeng , I have refreshed again. please review again

guoyuepeng:
@toyboxman 

could you fix the conflict?

